PR #11, opened 2015-01-13
BZ:1178090 - Administration: Revert button misnamed

nmirasch:


manstis:
If you change the "key" in the .properties file (like you have in the commit) you will also need to change the interface method [here](https://github.com/uberfire/uberfire-extensions/blob/master/uberfire-widgets/uberfire-widgets-core/uberfire-widgets-core-client/src/main/java/org/uberfire/ext/widgets/core/client/resources/i18n/CoreConstants.java#L155) and any use of the interface method elsewhere in the code-base.

manstis:
Also, if you change the key in one .properties file you will need to change the key in the other i18n .properties files.

PR #13, opened 2015-01-26
BZ:1014358 - Pagination and filtering on tables

nmirasch:
Added a new button to allow to change the page number of items

### DIFF
--- a/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/PagedTable.java
+++ b/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/PagedTable.java
@@ -16,13 +16,23 @@
 
 package org.uberfire.ext.widgets.common.client.tables;
 
+import com.github.gwtbootstrap.client.ui.Button;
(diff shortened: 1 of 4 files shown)


uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/PagedTable.java
```diff
@@ -45,39 +55,43 @@
     @UiField
     public UberfireSimplePager pager;
 
+    @UiField(provided = true)
+    public Button pageSizesSelector;
+
+    private boolean showPageSizesSelector = false;
+    private PopupPanel popup = new PopupPanel(true);
```
manstis: Please use BaseModal rather than PopupPanel.

We're trying to replace use of PopupPanel with BaseModal to improve (1) appearance (2) increase consistency.

nmirasch:
The PopupPanel is used to view the RadioButon options as a Dropdown button. It doesn't show a generic popup requesting or showing information to the user. The column picker is shown using the same PopupPanel

Salaboy:
merged in master

PR #16, opened 2015-03-17
Bug 1159254 - Asset metadata - commits don't show up

Rikkola:
Since I had to move few files from guvnor. Merging this will break guvnor/kie-wb-commons and so on. Once the PR has been reviewed I can do the merge and at the same time commit the changes needed for the other projects.

### DIFF
--- a/uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/version/PathResolverImpl.java
+++ b/uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/version/PathResolverImpl.java
@@ -61,10 +61,6 @@ public class PathResolverImpl
                 }
         }
 
-        public Path resolveDotFilePath(Path convert) {
-                return null;
-        }
-
         private Path getMainPath(Path path) {
                 return path.resolveSibling(util.getFileName(path).substring(1));
         }
--- a/uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/version/PathResolverImpl.java
+++ b/uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/version/PathResolverImpl.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2015 JBoss Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.uberfire.ext.editor.commons.backend.version;
+
+import java.net.URISyntaxException;
+import java.util.List;
+import javax.inject.Inject;
+
+import org.uberfire.java.nio.base.version.VersionRecord;
+import org.uberfire.java.nio.file.Path;
+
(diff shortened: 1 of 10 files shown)


uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/version/PathResolverImpl.java
```diff
@@ -0,0 +1,76 @@
...
+
+                        return mainPath;
+                } else {
+
+                        return path;
+                }
+        }
+
+        public Path resolveDotFilePath(Path convert) {
+                return null;
```
manstis: This looks a bit dangerous?

uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/version/VersionServiceImpl.java
```diff
@@ -48,47 +46,40 @@
...
-            return result;
+    @Override
+    public List<VersionRecord> getVersions(final Path path) {
 
-        } catch ( Exception e ) {
-            throw new RuntimeException( e );
+        try {
+            return versionRecordService.load(Paths.convert(path));
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
manstis: If you want the exception to propagate to the UI it might be better to use org.guvnor.common.services.backend.exceptions.ExceptionUtilities.handleException(e) as this returns a @Portable exception to the client which can then show a generic exception popup (if the client uses org.uberfire.ext.widgets.common.client.callbacks.DefaultErrorCallback).

uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/version/VersionServiceImpl.java
```diff
@@ -48,47 +46,40 @@
...
+
+            return convert(ioService.copy(path, target, REPLACE_EXISTING,
+                                          new CommentedOption(
+                                                  sessionInfo != null ? sessionInfo.getId() : "--",
+                                                  sessionInfo != null ? sessionInfo.getIdentity().getIdentifier() : "system",
+                                                  null,
+                                                  comment)));
+
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
manstis: Same as above.

uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/version/VersionServiceImpl.java
Rikkola: Would be nice. That class is in guvnor however.
I see that uberfire-wires has a copy&paste version of GenericPortableException and ExceptionUtilities. Maybe the exception utilities functionality should be moved into uberfire-extensions?

uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/version/PathResolverImpl.java
Rikkola: It is. It is also not used so killing it.

uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/version/VersionServiceImpl.java
manstis: Yes, that could be useful.

ATM the copy and paste is temporary (see https://github.com/uberfire/uberfire-extensions/blob/master/uberfire-wires/uberfire-wires-bpmn/uberfire-wires-bpmn-backend/src/main/java/org/uberfire/ext/wires/bpmn/backend/todo/ExceptionUtilities.java#L23).

PR #18, opened 2015-03-17
UF-174 : Move of GenericPortableException and ExceptionUtilities into ub...

Rikkola:
...erfire-extensions


manstis:
It's dangerous splitting a JAR between /client (GWT code) and /server (backend code). The "client" JARs are excluded from the WAR assemblies and hence the /server classes will not be available. I don't suppose you tried compiling kie-drools-wb's WARs and confirming this isn't going to cause a problem?

manstis:
IDK whether moving these to https://github.com/uberfire/uberfire-extensions/tree/master/uberfire-commons-editor/uberfire-commons-editor-backend is going to be better? We use these classes from our editor related code (in drools-wb etc) and use by Wires is for the same reason... just thinking aloud to avoid splitting a JAR across client-server (as we exclude -client JARs in our distributions).

Rikkola:
It is not that important where these end up. I just noticed there is a need for them in Uberfire. I will review this PR and see if the locations you suggested are better.

PR #19, opened 2015-04-01
REV: save user workbench preferences user in repo for workbecnh preferen...

bxf12315:
...ce setting menubar
@Salaboy  @manstis 
Please review it .

### DIFF
--- a/uberfire-widgets/uberfire-widgets-service/uberfire-widgets-service-backend/src/main/java/org/uberfire/ext/services/backend/preferences/UserWorkbenchPreferencesServiceImpl.java
+++ b/uberfire-widgets/uberfire-widgets-service/uberfire-widgets-service-backend/src/main/java/org/uberfire/ext/services/backend/preferences/UserWorkbenchPreferencesServiceImpl.java
@@ -1,0 +1,59 @@
+package org.uberfire.ext.services.backend.preferences;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import javax.inject.Named;
+
+import org.jboss.errai.bus.server.annotations.Service;
+import org.uberfire.backend.server.UserServicesBackendImpl;
+import org.uberfire.io.IOService;
+import org.uberfire.java.nio.file.Path;
+import com.thoughtworks.xstream.XStream;
+import org.uberfire.ext.services.shared.preferences.UserWorkbenchPreferencesService;
+import org.uberfire.ext.services.shared.preferences.UserWorkbenchPreferences;
+
+@Service
+@ApplicationScoped
+public class UserWorkbenchPreferencesServiceImpl implements
+        UserWorkbenchPreferencesService {
+    private static final String workbenchSettings = "workbench-settings";
+    private static final String userPreferences = "user-preferences";
+    @Inject
+    private UserServicesBackendImpl userServicesBackend;
+
+    @Inject
+    @Named("configIO")
+    private IOService ioServiceConfig;
+
+    private XStream xs = new XStream();
+
+    @Override
+    public void saveWorkbenchPreferences(UserWorkbenchPreferences preferences,
+            String identity) {
+        final Path preferencesPath = userServicesBackend.buildPath(identity,
+                userPreferences, workbenchSettings);
+        try {
+            ioServiceConfig.startBatch(preferencesPath.getFileSystem());
+            ioServiceConfig.write(preferencesPath, xs.toXML(preferences));
+        } finally {
+            ioServiceConfig.endBatch();
+        }
+
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


uberfire-widgets/uberfire-widgets-service/uberfire-widgets-service-backend/src/main/java/org/uberfire/ext/services/backend/preferences/UserWorkbenchPreferencesServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+    public UserWorkbenchPreferences loadWorkbenchPreferences(String identity) {
+        Path preferencesPath = userServicesBackend.buildPath(identity,
+                userPreferences, workbenchSettings);
+        try {
+            if (ioServiceConfig.exists(preferencesPath)) {
+                final String xml = ioServiceConfig
+                        .readAllString(preferencesPath);
+                return (UserWorkbenchPreferences) xs.fromXML(xml);
+            }
+        } catch (final Exception e) {
```
manstis: Swallowing exceptions without logging or reporting back to the UI is dangerous.. I'd prefer to see it logged or re-thrown as a RuntimeException.

uberfire-widgets/uberfire-widgets-service/uberfire-widgets-service-backend/src/main/java/org/uberfire/ext/services/backend/preferences/UserWorkbenchPreferencesServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+        return null;
```
manstis: Is it better to return an empty WorkbenchPreferences object than null? Otherwise we need to defensively check if the return value is null where-ever we choose to use preferences.

manstis:
Can you please use the code-style formatters we have available at: https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/tree/master/ide-configuration

manstis:
Please also consider how we may refactor the preferences service we have in kie-wb-common to use this new service: https://github.com/droolsjbpm/kie-wb-common/tree/master/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/preferences

Salaboy:
Besides @manstis corrections I would add that this PR should include the refactoring of the GridPreferenceServices so both services uses the same approach and share as much code as possible in a parent class.

manstis:
Finally(!) it might be better to have UserWorkbenchPreferences be a Map<String, String> of settings than explicit properties - just thinking when we want to increase the use of the service.

Salaboy:
Ok, based on the last comment from @manstis this seems a great opportunity to unify all these preferences services and standardise all the implementations so we can keep consistency between all of them.

manstis:
@Salaboy agreed.

@bxf12315 What we propose is no doubt a lot more work; but it's the best way to go.. if we don't stop the rot now, we may never stop the rot (rot being duplication instead of re-use).

uberfire-widgets/uberfire-widgets-service/uberfire-widgets-service-backend/src/main/java/org/uberfire/ext/services/backend/preferences/UserWorkbenchPreferencesServiceImpl.java
bxf12315: But my concern is  , how to handle the RuntimeException ?  Catch this exception in popup page, give some notification to user. than we just do noting in here. i means is if we can not get  preferences from git, we do not need do anything.

uberfire-widgets/uberfire-widgets-service/uberfire-widgets-service-backend/src/main/java/org/uberfire/ext/services/backend/preferences/UserWorkbenchPreferencesServiceImpl.java
bxf12315: got you mean, i need refer https://github.com/droolsjbpm/guvnor/blob/master/guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/config/AppConfigServiceImpl.java#L54, rigjt ?

bxf12315:
@manstis @Salaboy  , after leaning application preferences  framework,  i think refacing grid preference and user workbench preference to  application preferences  framework , this way is better.. But i find in the framework , we use java basic types store in git.  Grid and user workbench use custom class,. if we want merge them, we need refacing application preferences , let it support Configgroup and mapping Configgroup and our custom class.My question is can i refacing theses code ? it is relate to a lot of code .

PR #20, opened 2015-04-02
Add grid custom filters

nmirasch:
Adds the filter management and the dropdown list with defined filters to be applied in the displayed grid


Salaboy:
merged in master

PR #22, opened 2015-04-08
BZ-1206627: adapted to latest uberfire shutdown mechanism

porcelli:


manstis:
http://github.com/uberfire/uberfire-extensions/commit/83df5ebf3

PR #23, opened 2015-04-13
REV: create common preferences for workbench and grid

bxf12315:
@manstis @Salaboy 
Please review it.
thansk!


uberfire-widgets/uberfire-widgets-service/uberfire-widgets-service-api/src/main/java/org/uberfire/ext/services/shared/preferences/UserPreference.java
```diff
@@ -0,0 +1,26 @@
+package org.uberfire.ext.services.shared.preferences;
+
+import org.jboss.errai.common.client.api.annotations.Portable;
+
+@Portable
+public class UserPreference {
+    UserPreferencesType type;
```
manstis: Should this be "package" access level; private or protected may be more suitable.

uberfire-widgets/uberfire-widgets-service/uberfire-widgets-service-api/src/main/java/org/uberfire/ext/services/shared/preferences/UserPreference.java
```diff
@@ -0,0 +1,26 @@
...
+    
+    String preferenceKey ;
```
manstis: Same as above.

bxf12315:
Raise new PR

PR #28, opened 2015-04-22
 Add Filters on PagedTable As tabs

nmirasch:


manstis:
@salaboy, @nmirasch 

Shouldn't org.uberfire.ext.services.backend.preferences.UserDataGridPreferencesServiceImpl be deleted and use of org.uberfire.ext.services.backend.preferences.UserPreferencesServiceImpl be used instead now? (following on from @bxf12315's work)

Salaboy:
@manstis yes.. that's the idea.. I'm not sure  @bxf12315 will do that adaptation later .. as far as I know he was planing to do that.

bxf12315:
@Salaboy  @nmirasch 
As Salaboy said , we can use UserPreferencesServiceImpl instead of old code UserDataGridPreferencesServiceImpl. It is work.

PR #31, opened 2015-04-28
Extracting Layout Editor from Perspective Editor. Now perspective editor...

ederign:
... uses layout editor dnd infra structure.


manstis:
What's the impact of this on drools-wb-webapp, kie-(drools)-wb-webapp and presumably dashbuilder? Do they need changes to dependencies, compileSourceArtifacts and gwt.xml files?

ederign:
Michael for drools wb and kie-wb distribuitions apps is the PR's linked.

For dashbuilder, idk i'll confirm with david.

BTW this is a abstraction API of our former perspective-editor (a module inside runtime plugins). This was built to make it reusable for form builder.

The dashbuilder client is not inside this refactoring. I'll ask David to work on it.

uberfire-layout-editor/uberfire-layout-editor-client/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+      <artifactId>uberfire-widgets-commons</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.gwt.gwtmockito</groupId>
+      <artifactId>gwtmockito</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
```
manstis: Do we still need to do this in the new post-GWT2.7/Wildfly world?

uberfire-runtime-plugins/uberfire-runtime-plugins-backend/src/main/java/org/uberfire/ext/plugin/backend/PluginServicesImpl.java
```diff
@@ -615,44 +610,41 @@ public DynamicMenu getDynamicMenuContent( org.uberfire.backend.vfs.Path path ) {
                                 loadMenuItems( pluginName ) );
     }
 
+//    protected PerspectiveEditor makeDefaultPerspectiveEditor( final String pluginName ) {
```
manstis: Any reason you prefer to add commented code blocks rather than delete (as it's retained in the version history)? ;)

manstis:
Can you please explain the reason for the "generic" (i.e. I see you've done it in a lot of places) change:-

-        Button remove = new Button( "Remove" );
+        Button remove = GWT.create( Button.class );

manstis:
http://github.com/uberfire/uberfire-extensions/commit/5a6fe32d1

uberfire-layout-editor/uberfire-layout-editor-client/pom.xml
ederign: The gwt mockito scope?

uberfire-runtime-plugins/uberfire-runtime-plugins-backend/src/main/java/org/uberfire/ext/plugin/backend/PluginServicesImpl.java
ederign: I don't trust on git/github :)

Shame on me, I'll fix it on next PR.

uberfire-layout-editor/uberfire-layout-editor-client/pom.xml
manstis: Nope the <resources> section to include the source in the JAR (L#103-L#114).

ederign:
@manstis 
"Can you please explain the reason for the "generic" (i.e. I see you've done it in a lot of places) change:-

Button remove = new Button( "Remove" );
Button remove = GWT.create( Button.class );"

In order to work (mock) , GWT mockito require that the components use GWT.create instead of new.

I also think it's better to use GWT deffered binding for UI components.

BUT (there is always a but).

GWT Bootstrap Columns doesn't allow the use of default constructors (column is a required property).

And the big problem is that inside the construtor of Column, the frameworks uses for a assert GWT.create(ColumnSizeConfigurator.class).

If we do new Column(2) in a gwt mockito scope, it will automatically mock GWT.create(ColumnSizeConfigurator.class) and the default size of column will be the mock reply (default size == 0), making all column size assertions fails.

That's why we need to register the right provider like here: 

https://github.com/uberfire/uberfire-extensions/blob/master/uberfire-layout-editor/uberfire-layout-editor-client/src/test/java/org/uberfire/ext/layout/editor/client/dnd/DropColumnPanelTest.java#L34-L41

PR #32, opened 2015-04-30
Copy popup test

yurloc:
@mrietveld @ederign please review.

This is improved version of [CopyPopupTest](https://gist.github.com/yurloc/eb8bc92a6c4ee4731c72) from October 2014. I have separated CopyPopup view layer from the presentation logic (following MVP). The code has become more testable. I could avoid most of the hacks I complained about in this [gist](https://gist.github.com/yurloc/eb8bc92a6c4ee4731c72). Strict separation of widgets usage into view prevents any GWT.create() calls during test execution and so GwtMockito is not needed.

### DIFF
--- a/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopup.java
+++ b/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopup.java
@@ -13,31 +13,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.uberfire.ext.editor.commons.client.file;
 
-import com.github.gwtbootstrap.client.ui.TextBox;
-import com.github.gwtbootstrap.client.ui.constants.ButtonType;
-import com.github.gwtbootstrap.client.ui.constants.IconType;
-import com.google.gwt.user.client.Window;
 import org.uberfire.backend.vfs.Path;
-import org.uberfire.ext.editor.commons.client.resources.CommonImages;
-import org.uberfire.ext.editor.commons.client.resources.i18n.CommonConstants;
 import org.uberfire.ext.editor.commons.client.validation.Validator;
 import org.uberfire.ext.editor.commons.client.validation.ValidatorCallback;
-import org.uberfire.ext.widgets.common.client.common.popups.FormStylePopup;
-import org.uberfire.ext.widgets.common.client.common.popups.footers.GenericModalFooter;
-import org.uberfire.mvp.Command;
 
 import static org.uberfire.commons.validation.PortablePreconditions.*;
 
-public class CopyPopup extends FormStylePopup {
+public class CopyPopup implements CopyPopupView.Presenter {
 
-    final private TextBox nameTextBox = new TextBox();
-    final private TextBox checkInCommentTextBox = new TextBox();
+    private final CopyPopupView view;
+    private final Path path;
+    private final Validator validator;
+    private final CommandWithFileNameAndCommitMessage command;
 
-    public CopyPopup( final Path path,
-                      final CommandWithFileNameAndCommitMessage command ) {
+    public CopyPopup( Path path, CommandWithFileNameAndCommitMessage command, CopyPopupView view ) {
         this( path,
               new Validator() {
                   @Override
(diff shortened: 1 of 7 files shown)


uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopup.java
```diff
@@ -13,31 +13,40 @@
...
-import org.uberfire.ext.widgets.common.client.common.popups.FormStylePopup;
-import org.uberfire.ext.widgets.common.client.common.popups.footers.GenericModalFooter;
-import org.uberfire.mvp.Command;
 
 import static org.uberfire.commons.validation.PortablePreconditions.*;
 
-public class CopyPopup extends FormStylePopup {
+public class CopyPopup {
+
+    public interface View extends HasVisibility {
```
yurloc: Allows for testing the popup with a mock view, which avoids any GWT.create() calls during test execution so the test neither GWTTestCase nor GwtMockito are needed.

It also defines interactions between view and presenter, which can be easily verified.

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopup.java
```diff
@@ -46,72 +55,62 @@ public void validate( final String value,
                       callback.onSuccess();
                   }
               },
-              command );
+              command, view );
```
yurloc: View implementation supplied by user-code (real implementation) or test (mock implementation).

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopup.java
```diff
@@ -46,72 +55,62 @@ public void validate( final String value,
...
+                                  path );
+        this.command = checkNotNull( "command",
+                                     command );
+        this.view = checkNotNull( "view",
+                                  view );
+        this.view.getCopyButton().addClickHandler( new ClickHandler() {
+
+            @Override
+            public void onClick( ClickEvent event ) {
+                copy();
```
yurloc: Do the real work in a dedicated method. Separation between view-presenter glue boilerplate improves readability and testability.

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopup.java
```diff
@@ -46,72 +55,62 @@ public void validate( final String value,
...
+
+    public void show() {
+        view.show();
+    }
+
+    private void hide() {
+        view.hide();
     }
 
+    void copy() {
```
yurloc: May be called from a test without invoking the right click handler.

mrietveld:
Thank you @yurloc -- especially for the comments.

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopupView.java
```diff
@@ -0,0 +1,87 @@
...
+        checkInCommentTextBox.setTitle( CommonConstants.INSTANCE.CheckInComment() );
+        checkInCommentTextBox.setWidth( "200px" );
+        addAttribute( CommonConstants.INSTANCE.CheckInCommentColon(),
+                      checkInCommentTextBox );
+        hide();
+
+        GenericModalFooter footer = new GenericModalFooter();
+        this.copy = footer.addButton( CommonConstants.INSTANCE.CopyPopupCreateACopy(),
+                                      IconType.SAVE,
+                                      ButtonType.PRIMARY );
```
yurloc: No Command here. The click handler executing the command must stay in the presenter. Otherwise, if the we pass command here, the click handler is created by footer and so there is no way for the test to invoke the command because this code (including the footer with the click handler) will be mocked out.

uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/popups/footers/GenericModalFooter.java
```diff
@@ -47,6 +47,10 @@ public GenericModalFooter() {
         add( uiBinder.createAndBindUi( this ) );
     }
 
+    /**
```
mrietveld: +10!

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopupView.java
```diff
@@ -0,0 +1,87 @@
...
+        return copy;
+    }
+
+    @Override
+    public HasClickHandlers getCancelButton() {
+        return cancel;
+    }
+
+    @Override
+    public void handleInvalidFileName( String baseFileName ) {
```
yurloc: Call to this method can be verified with Mockito.

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopupView.java
```diff
@@ -0,0 +1,87 @@
...
+        Window.alert( CommonConstants.INSTANCE.InvalidFileName0( baseFileName ) );
```
yurloc: I failed to stub Window.alert() calls and mock the static instance of CommonConstants (without using PowerMock). Not sure if it is possible with (Gwt)Mockito or not.

uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/popups/BaseModal.java
```diff
@@ -36,18 +36,22 @@
...
+ * </ul>
  * <p/>
  * Furthermore this Modal provides:
- * - Automatic focus to the first Focusable widget in the body
- * - Automatic invocation of the first Button's ClickHandler where ButtonType==PRIMARY when <enter> is pressed
+ * <ul>
+ * <li>Automatic focus to the first Focusable widget in the body</li>
+ * <li>Automatic invocation of the first Button's ClickHandler where {@link ButtonType}==PRIMARY when &lt;enter&gt; is
+ * pressed</li>
+ * </ul>
```
yurloc: Just made it JavaDoc/IDE friendly.

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopupView.java
Rikkola: What I do is. I actually do keep the command in the view, but then pass the action to the presenter with a method that has a name that describes the action. This way the presenter does not care about click handlers and since the method name in the presenter interface describes the action that is taken, the test is more readable.

Just thought I mention what I do. Any client side code that is testable with MVP warms my heart ;)


Example:
https://github.com/droolsjbpm/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/pages/SummaryPageView.java#L32-32

Rikkola: Bad example link above. Looks like there is not test for it ;)

https://github.com/droolsjbpm/drools-wb/blob/master/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/test/java/org/drools/workbench/screens/testscenario/client/KSessionSelectorTest.java#L211-211

yurloc: @Rikkola thanks for those examples! If I understand that correctly, applying the advanced MVP should result in yurloc/uberfire-extensions@65e1dadf56111635747f892ce3cb4d7799fa0834.
Is [this](https://github.com/yurloc/uberfire-extensions/blob/65e1dadf56111635747f892ce3cb4d7799fa0834/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopupView.java#L52-L58) what you mean by keeping the command in the view?

Rikkola: Sorry missed you in IRC, but always happy to talk.

Yes that is what I meant. if you use presenter.onCopy() instead of adding a ClickHandler with HasClickhandlers interface for the copy button, the implementation ends up being more clean, but this is just my opinion and the way we do it in the guvnor/kie-wb-common/drools-wb/kie-wb. I don't think the MVP in Uberfire is as consistent and because of this I don't think the way it was done before was necessary wrong, it is still MVP. After trying out both ways I just found the use of Presenter interface over the HasHandler interfaces better.

yurloc: +1 I like the way with presenter interface better, too.

I wonder if Uberfire could re-use MVP interfaces from the [GWTP Project](http://dev.arcbees.com/gwtp/features/).

yurloc:
I am going to update this PR with advanced MVP suggested by Rikkola. View, presenter and test will all become cleaner and the (deprecated) methods `GenericModalFooter.addButton(String, Command[, IconType], ButtonType)` can be still used.

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopupView.java
manstis: IMO MVP is a conceptual pattern not tied to any specific implementation.
What benefit does GWTP bring (something IIRC we looked into and
discounted)? I did not understand their Proxy interface.

Sent on the move

yurloc:
Hehe, adding more tests is now a piece of :cake: :smile:

I consider this this PR ready.

yurloc:
@porcelli Can you handle this PR, please?

porcelli:
@yurloc mind squash commits before?

yurloc:
@porcelli sure!

yurloc:
@porcelli ready

porcelli:
https://github.com/uberfire/uberfire-extensions/commit/6d242f4df

PR #38, opened 2015-05-06
UF-180: Configurable Workbench

ederign:
- Fixing plugins missing (sometimes) in plugins editor nav list.

This happens randomly because in that lists is a screen and plugin with same name (L:215)

### DIFF
--- a/uberfire-runtime-plugins/uberfire-runtime-plugins-client/src/main/java/org/uberfire/ext/plugin/client/widget/navigator/PluginNavList.java
+++ b/uberfire-runtime-plugins/uberfire-runtime-plugins-client/src/main/java/org/uberfire/ext/plugin/client/widget/navigator/PluginNavList.java
@@ -212,8 +212,10 @@ public class PluginNavList extends Composite {
             //Sort Activities by Name. A TreeMap supports sorting on insertion by natural ordering of its keys
             final Map<String, Activity> activities = new TreeMap<String, Activity>( PLUGIN_NAME_COMPARATOR );
             for ( final Activity item : e.getValue() ) {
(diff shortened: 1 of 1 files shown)


uberfire-runtime-plugins/uberfire-runtime-plugins-client/src/main/java/org/uberfire/ext/plugin/client/widget/navigator/PluginNavList.java
```diff
@@ -212,8 +212,10 @@ public void setup( final Collection<Plugin> plugins ) {
             //Sort Activities by Name. A TreeMap supports sorting on insertion by natural ordering of its keys
             final Map<String, Activity> activities = new TreeMap<String, Activity>( PLUGIN_NAME_COMPARATOR );
             for ( final Activity item : e.getValue() ) {
-                activities.put( item.getName(),
-                                item );
+                if ( !thereIsAlreadyAPluginWithSameName( item, activities ) ) {
```
manstis: ...and what do you do if there is a plugin with that name already? Skip it, and then Users complain plugins are "disappearing"?

IMO we either need to prevent plugins from being created with an existing name - although this is going to be close to impossible if Users can create them with either (1) using the UI, (2) using a JSON file in the webapp's folder, (3) using a JSON file committed to the plugins.git repository; or find a solution to support activities with the sane name (but IDK if this is supported by the whole Plugins framework - or whether names need to be unique?).

manstis:
@ederign Thanks for your explanation yesterday.

PR merged: http://github.com/uberfire/uberfire-extensions/commit/e462e87c9

PR #40, opened 2015-05-12
JBPM-4625 Generation of JPA enabled models

wmedvede:
Small extensions to the properties editor widget in order to support customized fields.


ederign:
Seems ok to me. 
@manstis  wdyt?

ederign:
@wmedvede  thanks!

PR #43, opened 2015-06-10
Fix GWT warnings about Image resources being discovered on the ClassPath

manstis:


manstis:
http://github.com/uberfire/uberfire-extensions/commit/4ec086710

PR #47, opened 2015-06-29
BPMSPL-133: Tasklist should be refreshed automatically. Save refresh …

nmirasch:
…configuration in user preferences


Salaboy:
merged in master

PR #48, opened 2015-07-10
BZ-1241810: enforce cache control on HTTP headers (html is not enought)

porcelli:


manstis:
http://github.com/uberfire/uberfire-extensions/commit/e9e1e7bfd

PR #56, opened 2015-07-31
fix paged table page-size selector reloaded from preferences

nmirasch:
fix paged table page-size selector reloaded from preferences


Salaboy:
merged in master and 0.7.x

PR #58, opened 2015-08-12
Provide CopyPopup constructors that don't require view impl

yurloc:
In PR #32 I added constructor parameter which is an undesired API change. This PR returns the original constructors.


psiroky:
@porcelli could you please look into this as we need this to fix the build error(s) in kie-wb-common.

porcelli:
Sorry. I fixed it local but forgot to push; here's http://github.com/uberfire/uberfire-extensions/commit/150a2aec1

PR #60, opened 2015-08-21
Screen  context buttons on expanded Docks

ederign:


uberfire-simple-docks/uberfire-simple-docks-client/src/main/java/org/uberfire/client/docks/view/menu/MenuBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Inject
+    private AuthorizationManager authzManager;
+
+    @Inject
+    private User identity;
+
+
+    public Widget makeItem(final MenuItem item,
+                           boolean isRoot) {
```
Rikkola: Personally I would split this method to shorter methods. I'll merge anyway, not a cause for rejection.

ederign:
Merged https://github.com/uberfire/uberfire-extensions/commit/14888c859

ederign:
@Rikkola ( @porcelli ) I totally agree with our comment. This is a copy and paste of here (https://github.com/porcelli-forks/uberfire/blob/master/uberfire-workbench/uberfire-workbench-client-views-bs2/src/main/java/org/uberfire/client/views/bs2/listbar/ListBarWidgetImpl.java#L498).

I keep that in that way because I'll soon review that mechanism with @porcelli .

PR #61, opened 2015-08-27
upgrade to jboss-ip-bom 6.0.0.CR28

mbiarnes:
as uf-e inherits from jboss-ip-bom 6.0.0.CR28 which includes the version 
org.jgroups:jgroups:3.1.14.Final - but this version is not existing on public Nexus whe had to add
org.jgroups:jgroups:3.1.13.Final to UE-deps.
This had to be done if U-ext will upgrade to CR28 and will be there until we found a solution.


porcelli:
http://github.com/uberfire/uberfire-extensions/commit/42cd70bd3

PR #67, opened 2015-09-14
upgraded SNAPSHOT version

mbiarnes:
since 0.7.0.Final is out the 0.7.0-SNAPSHOT has to be upgraded to 0.7.1-SNAPSHOT


mbiarnes:
we have to upgrade to 0.7.2-SNAPSHOT as 0.7.1.Fibal is out

porcelli:
Close as this should be commited directly during build

PR #75, opened 2015-09-23
UF-228 - Configurable Workbench: Warning about unsaved changes without changing anything

ederign:
On Configurable workbench, when you open a SplashScreens/Screens/Editors even when you don't change anything, the warning about unsaved changes are displayed.

We have to set originalHash on Base editor when SplashScreens/Screens/Editors  are loaded because:

        @OnMayClose
        public boolean mayClose( Integer currentHash ) {
        if ( isDirty( currentHash ) ) {
            return baseView.confirmClose();
        } else {
            return true;
        }
    }

    public boolean isDirty( Integer currentHash ) {
        if ( originalHash == null ) {
            return currentHash != null;
        } else {
            return !originalHash.equals( currentHash );
        }
    }

### DIFF
--- a/uberfire-runtime-plugins/uberfire-runtime-plugins-client/pom.xml
+++ b/uberfire-runtime-plugins/uberfire-runtime-plugins-client/pom.xml
@@ -19,6 +19,18 @@
     <!-- dependencies added because of new illegal transitive dependency check -->
 
     <dependency>
+      <groupId>com.google.gwt.gwtmockito</groupId>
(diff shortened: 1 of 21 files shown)


ederign:
@csadilek  probably. Please hold on on this PR and I'll investigate that as soon as I finish my current task! Thanks!

ederign:
https://github.com/uberfire/uberfire-extensions/pull/81

Updated version

PR #76, opened 2015-09-23
Fixed UF-230: Adding/Removing docks can cause incorrect workbench layout 

ederign:
Adding/Removing docks needs to trigger resizing of the workbench layout so that other components can adjust their size accordingly.


csadilek:
master: ec01a67c9e96a1992ce111294be147f89bb8f3a8
0.7.x: 665c21185909073100139d227d692667191f6451

PR #77, opened 2015-09-24
[RFC] initial preferences API use

qmx:


qmx:
this depends on https://github.com/uberfire/uberfire/pull/184

manstis:
I know this is an RFC but I'd like to have seen removal of the "other" Preferences Services entirely, e.g. remove ```UserPreferencesService``` and use your new ```PreferenceStorage``` directly. While other "preferences services" exist I think there's a greater risk of people adding/amending to them in ways that circumvent your new service. Only my opinion :)

ederign:
+1 por @manstis

csadilek:
Closing since uberfire/uberfire#184 was closed.

PR #78, opened 2015-09-28
GUVNOR-2272: Post-BS3 migration: File Upload screen stopped working

manstis:
See https://issues.jboss.org/browse/GUVNOR-2272


csadilek:
Looks good to me. I've factored out the workaround for the GWT form encoding issue into a separate class/method and added a test to ensure we call it. This way we also have the beginning of some test coverage for the file upload editors. 

I've pushed this to my fork. If you agree with my changes I'd suggest cherry-picking them and squashing the two commits: https://github.com/csadilek/uberfire-extensions/commit/4e2490ffcb35115d67ea741d1fd611533d41dc3f

manstis:
@csadilek Your fork seems great; please proceed (I've added a couple of comments re: license headers in your fork).

csadilek:
Sure, I will add the licence headers but you'll have to do the proceeding so that the PR can automatically update ;). Simply cherry-picking and amending your commit should update this PR.

manstis:
@csadilek OK, all merged and updated.

There's a related PR for ```kie-wb-common``` https://github.com/droolsjbpm/kie-wb-common/pull/94.

This will need merging as well, if we're to avoid a broken build.

csadilek:
Thanks, @manstis! 

On master: 9dbbd2a621e917683b1928512ab388241a7b3dcb

PR #80, opened 2015-09-29
Add no argument constructor to client-side application-scoped bean

mbarkley:
This change helps prepare Uberfire for Errai 4.0, where normal scope beans will have to be proxiable.

### DIFF
--- a/uberfire-simple-docks/uberfire-simple-docks-client/src/main/java/org/uberfire/client/docks/UberfireDocksImpl.java
+++ b/uberfire-simple-docks/uberfire-simple-docks-client/src/main/java/org/uberfire/client/docks/UberfireDocksImpl.java
@@ -55,6 +55,10 @@ public class UberfireDocksImpl implements UberfireDocks {
     @Inject
     private Event<UberfireDockReadyEvent> dockReadyEvent;
 
+    // For proxying
+    public UberfireDocksImpl() {
+    }
+
     @Inject
     public UberfireDocksImpl( DocksBars docksBars ) {
         this.docksBars = docksBars;


manstis:
+1 to merge.

ederign:
My only concern about that is to put default constructors in everything for the sake of Proxy gods. This really makes me uncomfortable because the object is sometimes in a "invalid" state.

But I understand that this probably is a requirement.

My only question is: Can we have another option for this?

manstis:
@ederign good question.. but Errai is not alone, we have zero-parameter constructors all over the place for WELD.... I guess has anybody looked into whether the zero-parameter constructors need to be "public".. is "package scope" OK.. or even better private constructors?

ederign:
@mbarkley @csadilek is that possible?

csadilek:
@ederign It's a CDI requirement, really. Errai 4's CDI implementation is closer to the spec and behaves like WELD where before it didn't. The reason I haven't merged this yet is that we keep thinking about possible workarounds....but they are all just hacks.

csadilek:
@ederign I should also point out that the fact that all normal scoped beans can be trusted to be proxiable makes the design and implementation a lot cleaner and easier to understand.

psiroky:
But doesn't that effectively mean that you can not create immutable beans (unless they have zero fields)? From my point of view, that is quite bad...although it is already required directly by CDI spec.

Just my $0.02, as I don't know the details here.

csadilek:
@psiroky yes that's what it means. Here's the relevant chapter from the WELD docs: http://docs.jboss.org/weld/reference/latest/en-US/html_single/#_client_proxies

@ederign @psiroky The best solution around this, imo, is to use less normal scoped beans on the client. This just means less application scoped beans on the client as request scope etc. don't make sense there anyway. Getting rid of some singletons isn't a bad idea.

csadilek:
@mbarkley JSNI allows instantiating beans using their private constructor (::new()()). We could investigate that as a solution for the client-side (instantiate the proxies using JSNI). Then at least the no-arg constructor can be private and the beans can be effectively immutable (from Java code). //cc @psiroky @ederign

mbarkley:
@psiroky @csadilek Regarding immutable fields for normal scoped types, you can still use constructor injection to assign to immutable fields. But obviously it's annoying as you will have to write the boilerplate default constructor that assigns `null` to those fields.

@csadilek Regarding JSNI for using private constructors, I think that would be a good improvement, but it will still not help with the boilerplate part of the problem (assigning the final fields). Maybe for normal scoped classes with no default constructor the framework could just call the `@Inject` constructor with `null` for all parameters? Then classes would only have to define default constructors if they do work in their constructor (which I think is atypical?).
//cc @ederign

csadilek:
@mbarkley the only problem I see with this approach is that some of the classes do more with the constructor parameters than just assigning to a (final) field which could lead to weird (from the users perspective) NPEs, but I think adding both approaches would do it. If you add a private no-arg constructor it will be used for proxying, if it's not there we can try to construct the proxy using another constructor we find plus catch all exception and write a nice error message to tell the use how to prevent it (i.e. by adding a (private) no-arg constructor).

mbarkley:
Ok, I will attempt this change in Errai (allowing non-public constructors for proxying and using the `@Inject` constructors with `null` values when no default constructor is available).

I'm closing this PR since it may not be necessary after these changes.

PR #81, opened 2015-09-30
UF-228 - Configurable Workbench: Warning about unsaved changes without changing anything

ederign:
This is a updated version of: https://github.com/uberfire/uberfire-extensions/pull/75

Updates
======
Add a Refactoring removing duplicated logic from editors and moving to super class (as @csadilek requested).
Also adding coverage to load content method.

PR-75
=====
On Configurable workbench, when you open a SplashScreens/Screens/Editors even when you don't change anything, the warning about unsaved changes are displayed.

We have to set originalHash on Base editor when SplashScreens/Screens/Editors  are loaded because:

        @OnMayClose
        public boolean mayClose( Integer currentHash ) {
        if ( isDirty( currentHash ) ) {
            return baseView.confirmClose();
        } else {
            return true;
        }
    }

    public boolean isDirty( Integer currentHash ) {
        if ( originalHash == null ) {
            return currentHash != null;
        } else {
            return !originalHash.equals( currentHash );
        }
    }

### DIFF
--- a/uberfire-runtime-plugins/uberfire-runtime-plugins-client/pom.xml
+++ b/uberfire-runtime-plugins/uberfire-runtime-plugins-client/pom.xml
@@ -19,6 +19,18 @@
     <!-- dependencies added because of new illegal transitive dependency check -->
 
     <dependency>
+      <groupId>com.google.gwt.gwtmockito</groupId>
+      <artifactId>gwtmockito</artifactId>
+      <scope>test</scope>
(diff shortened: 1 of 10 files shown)


csadilek:
master: 870ce8c529accf240ab9949f2274f243e96c800c
0.7.x: 1237411981388f7af220ce977ee87396d85cc7d2

Thanks!

PR #84, opened 2015-10-01
BZ:1265600 - WAS: ClassNotFoundException: org.uberfire.ext.widgets.co…

nmirasch:
…mmon.client.tables.popup.DataGridFilterSummary

To avoid error message for using the annotation @portable


csadilek:
Thanks!

0.7.x: de0d95a46c422137a97f1271a4f1a92e4ad0f8d9

PR #85, opened 2015-10-14
UF-227: More dock resize fixes (see below)

csadilek:
This fixes the following two issues:

- When opening an editor with east docks (i.e. data modeller) the
editor's initial horizontal size is incorrect (it's too wide and
part of its content is hidden behind the docks)

- When switching to the source tab after expanding a dock, the
source view isn't properly resized and is limiting its horizontal
size by the width of the dock that is no longer displayed


csadilek:
This also needs to be cherry-picked to 0.7.x.

ederign:
Merged! Thanks!
master: https://github.com/uberfire/uberfire-extensions/commit/93ebf904f3b90d2ad62f3e6ab0886520257d5eac
0.7.x: https://github.com/uberfire/uberfire-extensions/commit/d6e84248884077a30efe84ee065070139c276495

PR #86, opened 2015-10-14
BZ-1195976 - [MetaData - Versions] Restoring an old version, doesn't change the Restore Button to Save

wmedvede:


Rikkola:
@wmedvede do you want me or @ederign? Or was I just CC'd?

wmedvede:
Sorry Toni, I initially assigned it to you and then I saw Eder had some
other stuff to merge into uberfire-extensions too and reassigned to him

2015-10-14 12:58 GMT+02:00 Toni Rikkola <notifications@github.com>:

Rikkola:
No problemo.

ederign:
@wmedvede could you please add tests to this PR? Thanks!

Rikkola:
Unit tests get tricky since SaveButton might throw the GWT.create exception. Most practical way would be to GwtMock the SaveButton.
Other option is to refactor SaveButton to MVP. This is worth it if the "implements MenuCustom<Widget>" in SaveButton can be changed to "implements MenuCustom<IsWidget>".

wmedvede:
Since it's the same case as for the "VersionMenuDropDownButton" I've basically used same approach. In this way we can have the same approach for the test too. So the test is now working.

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/history/VersionRecordManager.java
```diff
@@ -263,6 +265,14 @@ public void callback( List<VersionRecord> records ) {
...
     }
 
+    private void onRestore( @Observes RestoreEvent restore ) {
+        if ( getCurrentPath() != null &&
+                getCurrentPath().equals( restore.getPath() ) &&
+                saveButton != null ) {
+            saveButton.setTextToSave();
+        }
+    }
+
```
csadilek: This new method would've been a good candidate for a unit test ;).

PR #88, opened 2015-10-14
UF-238: Configurable Workbench: Drag and drop does not work on IE11

ederign:
We used to use the format parameter of HTML 5 native DND  event.setData( format, data) to specify which type of component are on drag and drop.

     event.setData( LayoutDragComponent.class.toString(), type.getClass().getName() );

However, this doesn't work on ie11 because it requires that the format parameter is exactly the String "text".

    public static final String FORMAT = "text";
     ...
    event.setData( DndData.FORMAT,  DndData.generateData( type ) )

### DIFF
--- a/uberfire-layout-editor/uberfire-layout-editor-client/src/main/java/org/uberfire/ext/layout/editor/client/dnd/DndData.java
+++ b/uberfire-layout-editor/uberfire-layout-editor-client/src/main/java/org/uberfire/ext/layout/editor/client/dnd/DndData.java
@@ -1,0 +1,59 @@
+/*
+* Copyright 2015 JBoss Inc
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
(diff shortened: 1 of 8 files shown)


ederign:
@manstis if I recall correctly, you use html5 drag and drop somewhere in workbench right? take a look at this.

ederign:
This also needs to be cherry-picked to 0.7.x.

manstis:
Sweet, so we have to encode everything in the value ;)

A personal preference would been to have also encoded "source" into the value to avoid the repeated use of ```isInternalDragComponent( String eventType )``` , e.g. have ```prepareData(...)``` do something like ```eventType + SEPARATOR + eventData + SEPARATOR + eventSource``` but otherwise I'm happy to merge (your code formatting is duff though....)

ederign:
@manstis pr updated. thanks for the review

csadilek:
Obrigado!

master: e126e5c2192ac7c49604dc040e644b9963b4d796
0.7.x: 8d2acf46f79721f98f33455cf17a06c04aef8911

PR #91, opened 2015-10-19
BZ1269333: Result of searching assets by Business Central includes duplicate records

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1269333

The issue was caused by ```LuceneIndexEngine``` writing "full text" index entries for all indexers; the default and additional ```Indexer``` implementations in ```kie-wb-common```. We now restrict creation of the "full text" index entries to the default indexing.


manstis:
See https://github.com/droolsjbpm/kie-wb-common/pull/113 too

manstis:
master and 0.7.x please

PR #94, opened 2015-10-21
Removed side margins from table components

cristianonicolai:
Most use cases only need top and bottom margins. Side margins often cause extra css to override and remove it.


manstis:
+1 to merge.

csadilek:
Thanks!

master: cc3dd9a0a3c386e8736eb19431ef96edc3a36c86

PR #99, opened 2015-10-28
Bug 1273036 - Strange behaviour of paginations in Overview Tab

Rikkola:
The bug is only in 0.7.x. Move to PatternFly fixed it for master.
Didn't do any tests, because those would require a rewrite to MVP and in the end the fix is nice and clear.

@mbiarnes needs a commit so he can cherry-pick, but he might not have be able to. The builds have already started and this bug is not enough to make him restart. 
This branch is probably frozen so we should not merge this in, unless @mbiarnes rebuilds. Leaving this here for review and for @mbiarnes to pick up if he can.

The ticket has a the changes in a patch file in case there is ever a need to fix this for 0.7.x.

### DIFF
--- a/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/PagedTable.java
+++ b/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/PagedTable.java
@@ -70,6 +70,7 @@ public class PagedTable<T>
         this.pageSize=pageSize;
         this.dataGrid.setPageSize( pageSize );
         this.pager.setDisplay( dataGrid );
+        this.pager.setPageSize( pageSize );
         setShowPageSizesSelector( showPageSizesSelector );
         createPageSizesListBox( 5, 20, 5 );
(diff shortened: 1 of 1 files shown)


csadilek:
I don't think we should rush this fix in, if it's agreed that it's not important for this release. We'd also need to write a test for it. The latter should be possible without the MVP refactoring as we could simply verify that the constructor is invoking setPageSize. Also, shouldn't the constructors be reused with calls to this() instead of repeating all their init logic?

Rikkola:
The constructors contents are redundant and can be cleaned. I tried to do a change with the clearest change in case we can rush it in. So I tried to change as little as possible. 
Storing the page size in a field variable, and other things that the class does, do not feel right, but we can clean this up if we ever need to. Probably we will never touch this branch again.

csadilek:
I understand, but let's be strict on ourselves. If we want it fixed, let's fix it properly (with refactoring and tests) on master. This will be better in the long run.

Rikkola:
The PatternFly upgrade fixed this for master. So no worries.

And not complaining about being more strict. I think we do better in the long run if we take that route.

Closing this now since this is not needed.

PR #100, opened 2015-10-30
BZ1275325: WAS: indexing fails with ContextNotActiveException

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1275325

Implementation of EJB was flawed; it was neither an "no interface" bean, nor a "local interface" bean.


csadilek:
0.7.x: e7df6e18d6a3a41989d057ce41d453657d9fdac7

PR #108, opened 2015-11-11
BZ:1261052 - Console: various localized perspective listings does not…

nmirasch:
… work.

### DIFF
--- a/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/ColumnMeta.java
+++ b/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/ColumnMeta.java
@@ -25,7 +25,7 @@ public class ColumnMeta<T> implements Comparable {
     private boolean visible = true;
     private int position = -1;
     private boolean extraColumn;
-    
+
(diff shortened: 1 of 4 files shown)


uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/ColumnMeta.java
```diff
@@ -55,6 +58,12 @@ public ColumnMeta(Column<T, ?> column, String caption,boolean visible, boolean e
         this.extraColumn = extraColumn;
     }
 
+    private void intiDataStoreName(String caption){
+        if(column.getDataStoreName()==null){
```
cristianonicolai: I don't think the ColumnMeta class should be responsible for this. Is actually the ColumnPicker that needs to understand whether to use the caption or dataStoreName. I think a better approach would be to check if  getDataStoreName() is null in the ColumnPicker.

nmirasch:
Moved the datastorename control to columnPicker

csadilek:
+1 and good work on the tests!

0.7.x:acfd90743df5ce49b1e2c6105f72fe85a64d34aa

@nmirasch please don't forget master ;).

PR #110, opened 2015-11-12
BZ1269333: Result of searching Assets in Business Central's 'Full Text Search' includes duplicate records

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1269333

The issue is "additional indexers" also create index entries for "full text" (but with a different ID to the "default indexing"). Consequentially when performing a "full text" search two records were matched: one created by "default indexing" and the other by the "additional indexers".

Fixed and added new test. Existing tests were refactored to use a new common base class to avoid duplication of the setup/dispose etc code.


manstis:
Related PR: https://github.com/droolsjbpm/kie-wb-common/pull/134

manstis:
@porcelli since you commented on the BZ care to review this?

porcelli:
+1

csadilek:
Thanks!

master: 90c4901f4082fb607b285d1491bd61ee28c6dc48

PR #116, opened 2015-11-12
Bug 1273036 - Strange behaviour of paginations in Overview Tab

Rikkola:
I'm really really not a fan of the makeDataGrid methods in SimpleTable. But the version of GwtMockito we use currently does not support stubbing of the DataGrid. So the options were to either use a protected methods that can be overwritten in the test or upgrade gwtmockito just for this module (not sure if even the latest supports it, but the snapshot does).

### DIFF
--- a/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/PagedTable.java
+++ b/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/PagedTable.java
@@ -67,33 +67,21 @@ public class PagedTable<T>
 
     public PagedTable( final int pageSize ) {
         super();
-        this.pageSize=pageSize;
-        this.dataGrid.setPageSize( pageSize );
-        this.pager.setDisplay( dataGrid );
-        setShowPageSizesSelector( showPageSizesSelector );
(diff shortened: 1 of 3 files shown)


manstis:
+1 to merge.

manstis:
I think this represents more than a reasonable compromise. The topic of the BZ has been fixed and tested together with some of the refactoring requested by Christian.

ederign:
@Rikkola what is the issue with mocking DataGrid?

csadilek:
Yes, agreed +1. Esp. as this fix is only for 6.3.x. Merging...

csadilek:
Thanks!

0.7.x: 603b966dfecb14fc5eb812139f72308664bbc0c9

Rikkola:
@ederign 
The DataGrid constructor does the following and getWidget() kept returning a mock that could not be cast to HeaderPanel.
``` 
public DataGrid(int pageSize, DataGrid.Resources resources, ProvidesKey<T> keyProvider, Widget loadingIndicator) {
        super(new HeaderPanel(), pageSize, new DataGrid.ResourcesAdapter(resources), keyProvider);
        this.headerPanel = (HeaderPanel)this.getWidget();
        ...
}
```

There is probably a workaround that I'm missing. Best way would be to stub DataGrid, but the support for that is not in the gwtmockito version we have now.

PR #118, opened 2015-11-13
license updates

porcelli:


ederign:
+1

@psiroky there is a way to include licenses verification in our CI?

psiroky:
@ederign I think it should be somehow possible using one the maven license plugins. I haven't really looked into that in more detail.

csadilek:
master: 2de657fea45fa0669d65876f734f305745b69000

PR #121, opened 2015-11-13
Enhance CacheHeadersFilter to allow for caching of .cache.* files

csadilek:
This makes the compiled permutations (*.cache.js) cacheable :) (because they are). When refreshing [1] or revisiting the host page, we will now get a 304 (Not Modified) and these files will be served from cache. Note that every new compile generates the *.cache.js permutation files with different hashes. This is on purpose so they are downloaded once after every compile/redeployment and we'll never end up with a cached version when a new one should be loaded.

This should significantly improve the loading time of our workbench over slow network connections. Of course, the host page and the *.nocache.js files are still not cached and downloaded every time.

More details here [2], expiry date was set to one year in the future as recommended by [3].

[1] regular refresh NOT ctrl->shift->r or any other cache invalidating reload
[2] https://www.mnot.net/cache_docs/
[3] https://www.ietf.org/rfc/rfc2616.txt

### DIFF
--- a/uberfire-security/uberfire-servlet-security/src/main/java/org/uberfire/ext/security/server/CacheHeadersFilter.java
+++ b/uberfire-security/uberfire-servlet-security/src/main/java/org/uberfire/ext/security/server/CacheHeadersFilter.java
@@ -17,15 +17,22 @@
 package org.uberfire.ext.security.server;
 
 import java.io.IOException;
+import java.util.Calendar;
+
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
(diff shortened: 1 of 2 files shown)


csadilek:
@manstis @mswiderski as discussed. Please take a look ;).

manstis:
Looks good.

+1 to merge.

mswiderski:
+1

csadilek:
@porcelli please merge ;).

porcelli:
https://github.com/uberfire/uberfire-extensions/commit/9b91985877bdceb76a1edd79af2fc368997e4056

PR #123, opened 2015-11-17
BZ-1232000: if lucene parser returns an empty query - we use widlcard to

porcelli:
return possible results. Usually this happens when you try to search
with a single character.


porcelli:
https://bugzilla.redhat.com/show_bug.cgi?id=1232000

manstis:
Wozza! By the way do you actually have a test for an empty term? Looking again I could not see one.

manstis:
I'm not questioning your fix - well I am actually ;-) - but is the cause that "a" is a stop word filtered by the default analyser used for full text.. so in essence the "full text" entry stored in the index is empty?

PR #124, opened 2015-11-18
BZ-1232000: removing the use of stop words from default analyzer -

porcelli:
default analyzer is used to index the fullText field. Now indexing  and
queries works as expected with stop words like "a" or "the"


manstis:
+1 to merge.

csadilek:
Thanks!

master: 2193ff5a3f8c1a37053e517379cba46cf9bf740c

PR #125, opened 2015-11-18
Enhance CacheHeadersFilter to allow for caching of .cache.* files

csadilek:
Cherry picked from master for (newly created) https://bugzilla.redhat.com/show_bug.cgi?id=1283304

This makes the compiled permutations (*.cache.js) cacheable :) (because they are). When refreshing [1] or revisiting the host page, we will now get a 304 (Not Modified) and these files will be served from cache. Note that every new compile generates the *.cache.js permutation files with different hashes. This is on purpose so they are downloaded once after every compile/redeployment and we'll never end up with a cached version when a new one should be loaded.

This should significantly improve the loading time of our workbench over slow network connections. Of course, the host page and the *.nocache.js files are still not cached and downloaded every time.

More details here [2], expiry date was set to one year in the future as recommended by [3].

[1] regular refresh NOT ctrl->shift->r or any other cache invalidating reload
[2] https://www.mnot.net/cache_docs/
[3] https://www.ietf.org/rfc/rfc2616.txt

### DIFF
--- a/uberfire-security/uberfire-servlet-security/src/main/java/org/uberfire/ext/security/server/CacheHeadersFilter.java
+++ b/uberfire-security/uberfire-servlet-security/src/main/java/org/uberfire/ext/security/server/CacheHeadersFilter.java
@@ -1,15 +1,22 @@
 package org.uberfire.ext.security.server;
 
 import java.io.IOException;
+import java.util.Calendar;
+
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
 import javax.servlet.FilterConfig;
 import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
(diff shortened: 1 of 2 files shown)


csadilek:
@mswiderski @manstis as discussed, we were asked to cherry-pick the cache improvements to 0.7.x/6.3.x.

manstis:
+1 to merge.

csadilek:
@krisv can you quickly confirm that we want this in the 6.2 release so that @porcelli can merge. ;)

porcelli:
@krisv ?

porcelli:
https://github.com/uberfire/uberfire-extensions/commit/21e205350b4af661bd9a278f2628e98e2d120f94

PR #134, opened 2015-11-27
Fix display name of the uberfire-apps project

yurloc:
The previous name was result of a copy&paste error and it was a source of confusion when navigating project structure in an IDE.


manstis:
+1 to merge.

csadilek:
Thanks.

master: b43e0f1e0a3b165ff3bc345d2e89d9eea342514b

PR #136, opened 2015-11-27
Fix bounding coordinates after resizing rectangular container

yurloc:
The X and Y coordinates of the bounding were doubled compared to the
container position.


manstis:
+1 to merge (this is as good as deprecated though, as all of Wires Scratch Pad is to be replaced by the work being done by @qmx and @romartin for the jBPM Designer replacement. The new works uses new functionality in Lienzo - this Scratch Pad served as a PoC).

csadilek:
master: daf42d6802417427d29552caae5976851b190f1a

PR #142, opened 2015-12-03
BZ1192831: User with no privileges for Repository can view and modify assets in that Repository

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1192831#c38

If no ```ClusterSegments``` are provided we searched *all* ```ClusterSegments``` leading to searches when the User is not authorised to access *any* Repository actually returning results from *all* Repositories. Making this change to uberfire-extensions is my preferred option as it ensures all current and future scenarios are covered (i.e. *all* searching in KIE Workbench should be scoped to the Repositories to which the User has access).

@porcelli @csadilek @ederign Thoughts please!


manstis:
* DO NOT MERGE * THIS IS FOR REVIEW AT THE MOMENT.

csadilek:
Looks good to me. This way you can differentiate between no segments, some segments and all segments using the same method. It's then also using the same logic as attribute filtering (no attributes provided -> empty list, no segments provided -> empty list). +1.

manstis:
DO NOT MERGE YET - FINAL TESTING IN PROGRESS

csadilek:
@porcelli objections? ;)

porcelli:
+1

csadilek:
Thanks.

master: 5b509dd066552d4a033e5a581d6d1c8330f39b1b

PR #145, opened 2015-12-04
BZ1288132: Same 'Completed indexing' message 3 times in the server log

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1288132

Batch Indexing was being flagged as "started" in the asynchronous handler, which itself may not have been scheduled by the JVM to run until *after* additional requests for the ```FileSystem``` had been received. Each request checks if the ```FileSystem``` has been indexed and, if not, schedules a batch index. This change moves flagging the batch index as "started" from the asynchronous handler to where the request is made. This prevents multiple batch indexing on successive requests for an (unindexed) ```FileSystem```.


porcelli:
+1 (thx for fix my sh*t)

csadilek:
Thanks.

master: 7d6764e69434756f22c54f965cbbb8ccfff943ee

PR #147, opened 2015-12-07
Use POMs directly from uberfire to reduce maintenance efforts

psiroky:
* ip-bom is used in uberfire, so when upgrading we only
   need changes there and not in this repo
 * both uf and uf-ext are always released together, so it
   is fine that they uf-ext depends on SNAPSHOTs of uf


psiroky:
We will probably need to wait till `jboss-integration-platform-parent:6.0.0.CR31` is synced into central, otherwise the build fails.

I will re-trigger the Jenkins build once the POM is synced, please don't merge until then.

psiroky:
Ready for review.

csadilek:
+1

csadilek:
Thanks.

master: 03bdc3a14651a5222270bdbf37af4f2d778d7377

PR #151, opened 2015-12-17
Avoid running concurrent batch indexing threads for the same FS

csadilek:
This fixes the intermittent test failures of BatchIndexConcurrencyTest [1]
and improves the fix for BZ-1288132 [2]

[1] https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/uberfire-extensions/186/org.uberfire$uberfire-metadata-commons-io/testReport/org.uberfire.ext.metadata.io/BatchIndexConcurrenyTest/testForSingleBatchIndexExecution/
[2] https://github.com/uberfire/uberfire-extensions/commit/7d6764e69

### DIFF
--- a/uberfire-metadata/uberfire-metadata-commons-io/src/main/java/org/uberfire/ext/metadata/io/BatchIndex.java
+++ b/uberfire-metadata/uberfire-metadata-commons-io/src/main/java/org/uberfire/ext/metadata/io/BatchIndex.java
@@ -16,6 +16,13 @@
 
 package org.uberfire.ext.metadata.io;
 
+import static org.uberfire.commons.validation.PortablePreconditions.checkNotNull;
+import static org.uberfire.java.nio.file.Files.walkFileTree;
+
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


manstis:
+1 to merge.

csadilek:
@manstis just waiting to see if jenkins is actually happy with this. would've asked for your review after. I hope I am not missing anything but I take your plus one as a good sign ;).

manstis:
I guess the only concern would be runAsync() should be synchronized...

csadilek:
@manstis yes, I had used a concurrenthashmap first before switching to identity....I will update to use a thread-safe set.

manstis:
IMO it's not the indexPending.add(..) that'll need to be thread safe but the invocation of SimpleExecutor... so still suspect it should be the method that is synchronized. Happy to be told otherwise though!

csadilek:
@manstis I would disagree with that because we're first checking the set and only then invoking the executor.

manstis:
To be honest, I'm no concurrency guru, so can accept your comment. Personally I still see potential for concurrency issues between the read and write... but if you're confident CopyOnWriteArraySet fits... then +1 to merge.

manstis:
If there was an atomic getAndSet then I'd agree whole heartedly :-)

csadilek:
@manstis I don't know what you mean. I think you might be misinterpreting what I am trying to do here. It's probably best to chat about it.

csadilek:
Closing this unmerged. The more I think about it the less I like fixing this on that level.

PR #152, opened 2015-12-18
Avoid running concurrent batch indexing threads for the same FS

csadilek:
This fixes the intermittent test failures of BatchIndexConcurrencyTest [1]
and improves the fix for BZ-1288132 [2]

[1] https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/uberfire-extensions/186/org.uberfire$uberfire-metadata-commons-io/testReport/org.uberfire.ext.metadata.io/BatchIndexConcurrenyTest/testForSingleBatchIndexExecution/
[2] 7d6764e

### DIFF
--- a/uberfire-metadata/uberfire-metadata-commons-io/src/main/java/org/uberfire/ext/metadata/io/BatchIndex.java
+++ b/uberfire-metadata/uberfire-metadata-commons-io/src/main/java/org/uberfire/ext/metadata/io/BatchIndex.java
@@ -150,6 +150,8 @@ public final class BatchIndex {
                                                     attrs );
 
                                       if ( !file.getFileName().toString().startsWith( "." ) ) {
+                                          
+                                          LOG.debug( "Indexing " + file.toUri() );
 
                                           //Default indexing
                                           for ( final Class<? extends FileAttributeView> view : views ) {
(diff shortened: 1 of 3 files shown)


csadilek:
@manstis I was able to enhance the test to make the issues reproducible and fixed it. The fix is almost as discussed, but I am not merging `index` and `indexIfFresh` because that caused another issue. Turns out `newFileSystem` was not interfering with `getFileSystem` after all (the indexing started by newFileSystem wouldn't have run as there are no root dirs). Therefore merging them would cause no batch index task to run. Anyways, this should fix the issue and if not we have more logging.

manstis:
So we synchronized the method after all ;-)

+1 to merge.

csadilek:
@manstis at least now the test is covering it and it's clear why it has to be synchronized (it's also a different method). so, something good came out of the discussions ;).

etirelli:
Merged.

csadilek:
@manstis @psiroky this undoubtedly fixed one concurrency issue, but unfortunately BatchIndexConcurrencyTest still fails on Jenkins: https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/uberfire-extensions/189/org.uberfire$uberfire-metadata-commons-io/

So, there's more involved and we need to come back to this.

PR #155, opened 2015-12-21
BZ:1283503 - Task list: loading animation displayed indefinitely unti…

nmirasch:
…l the mouse is moved


csadilek:
+1, I will merge this as it fixes an urgent problem, but we need to revisit this. What are the events that need to be executed before the onShow logic and why?

csadilek:
012d80040972735c57a30153f90474d9264778a4

PR #158, opened 2015-12-30
BZ-1197649 - Validation inline on copying and renaming plugins.

paulovmr:
The only thing left to resolve the BZ-1197649 ticket is to change the validation message accordingly to the ticket description, which I will do soon. Only making the PR now to avoid conflict in the meantime.

I also did a refactoring on the rename popup to separate the view from the presenter. Now this popup is very similar to the copy popup. They can be unified in the future much easier.

### DIFF
--- a/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/RenamePopupViewImpl.java
+++ b/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/RenamePopupViewImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 JBoss, by Red Hat, Inc
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,7 +30,7 @@ import org.uberfire.ext.widgets.common.client.common.popups.footers.GenericModal
 import org.uberfire.mvp.Command;
 
 @Dependent
-public class CopyPopupViewImpl extends FormStylePopup implements CopyPopupView {
+public class RenamePopupViewImpl extends FormStylePopup implements RenamePopupView {
 
     private final TextBox nameTextBox = new TextBox();
     private final TextBox checkInCommentTextBox = new TextBox();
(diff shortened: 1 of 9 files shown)


paulovmr:
jenkins, retest this

paulovmr:
jenkins, retest this

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopupViewImpl.java
```diff
@@ -85,6 +99,9 @@ public String getCheckInComment() {
 
     @Override
     public void handleInvalidFileName( String baseFileName ) {
-        Window.alert( CommonConstants.INSTANCE.InvalidFileName0( baseFileName ) );
+        nameFormStyleItem.getFormGroup().setValidationState( ValidationState.ERROR );
+
+        // TODO : Change text of the validation message to "Invalid plugin name."
```
manstis: Following our discussion this is correct if we're to have consistency across all rename/copy actions for all asset types.

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/RenamePopupViewImpl.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public String getCheckInComment() {
+        return checkInCommentTextBox.getText();
+    }
+
+    @Override
+    public void handleInvalidFileName( final String baseFileName ) {
+        nameFormStyleItem.getFormGroup().setValidationState( ValidationState.ERROR );
+
+        // TODO : Change text of the validation message to "Invalid plugin name."
```
manstis: Same here; re: error message for invalid asset names.

manstis:
You should check guvnor, kie-wb-common and drools-wb compile with the change to ```FormStyleLayout``` as I am pretty sure (at least) one of them used the index returned from ```addAttribute(...)``` - which now returns ```FormStyleItem```.

paulovmr:
Thanks @manstis . About the TODOs, I will leave that way until we have a answer to my comment on the ticket (https://bugzilla.redhat.com/show_bug.cgi?id=1197649#c4).

Depends on:
https://github.com/droolsjbpm/drools-wb/pull/105
https://github.com/droolsjbpm/guvnor/pull/210

manstis:
@paulovmr I'd still prefer the ```TODO's``` to be changed to reflect they're more "might have to do" rather than "outstanding thing to change". Sorry. IMO they make it look like the PR is incomplete and there remains something to fix - I'd at least reference the BZ (IDK something like "This message may need changing for different asset types; see https://bugzilla.redhat.com/show_bug.cgi?id=1197649#c4")

Otherwise +1 to merge.

paulovmr:
@manstis Ok, totally agree with you. I already made the changes. Thanks!

manstis:
+1 to merge (along with the other two referenced PRs).

csadilek:
master: 7cf337606b4f1356279e7117d1d307fad2197c70

PR #159, opened 2015-12-30
Additional fixes for BatchIndexConcurrencyTest (details below)

csadilek:
- Default indexing thread for created test file no longer
interfering with the indexing threads under test

- Fixed indexFresh check to account for the fact that
an index could have been created (!=null) but the
corresponding indexing job is still in progress in which
case we also don't want to start another indexing thread

- These fixes are in addition to the earlier fix that made
indexIfFresh synchronized


csadilek:
@manstis please take a look next week.

manstis:
+1 to merge.

psiroky:
@csadilek thanks for looking into this and fixing it!

PR #163, opened 2016-01-07
BZ-1197644: Configurable Workbench: Exception is displayed after copying or renaming plugin

paulovmr:


csadilek:
@paulovmr looks good. I think we should give the "fileAlreadyExists" case a separate error message as it's misleading to claim the name is not valid in this case. wdyt?

paulovmr:
@csadilek i agree. Only did it that way because the ticket described the error as "It should be same as in case of illegal name when you try to create new plugin". Should I change it then?

csadilek:
@paulovmr Yes, I think we have to change this. This affects renaming/copying functionality in the workbench i.e. when you rename/copy an asset/file, where before you would get a specific error [1], telling you that the file already exists.

[1] https://github.com/uberfire/uberfire-extensions/blob/master/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/callbacks/DefaultErrorCallback.java#L48

paulovmr:
@csadilek Done.

csadilek:
+1

csadilek:
master: a3eb059bec04a5152c8779e5d28eb223ace610b2

PR #171, opened 2016-01-12
UF-258: Plugins can be copied or renamed to non-editable plugin names

paulovmr:


ederign:
+1
@csadilek I've already build and test locally. Without "-Denforcer.skip=true" i got this error (not related with the PR)

[INFO] --- maven-enforcer-plugin:1.4:enforce (ban-unwanted-logging-deps) @ uberfire-runtime-plugins-client ---
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.BannedDependencies failed with message:
Found Banned Dependency: commons-logging:commons-logging:jar:1.1.1
Use 'mvn dependency:tree' to locate the source of the banned dependencies.

manstis:
Petr emailed (IIRC) BSIG about logging implementations being included
incorrectly in dependencies or transient dependencies and how they should
be handled. This looks like such an occurrence.

ederign:
It was a old dep only in my machine
[INFO] |  |     \- org.apache.httpcomponents:httpclient:jar:4.3.6:compile
[INFO] |  |        +- org.apache.httpcomponents:httpcore:jar:4.3.3:compile
[INFO] |  |        \- commons-logging:commons-logging:jar:1.1.1:compile

after mvn clean install -U it's gone. We are good to go.

manstis:
Yay

csadilek:
@paulovmr Please update the commit message though because it's a duplicate (see https://github.com/uberfire/uberfire-extensions/commit/45919d49e9d3eca9c5b70dab3da460a905362213). You can add additional details on subsequent lines so we can more easily differentiate when browsing history.  /cc @ederign

paulovmr:
@csadilek Ok, done.

csadilek:
Thanks.

master: 9a49599ce03cf2d6a70ca5127031041ba2f1fe65

PR #173, opened 2016-01-13
UF-265: User System Management: Group filtering and role assignment support

romartin:
- Added group filtering support in order to exclude the roles registered in UF
- Added service and user interface components to assign both groups and roles
- Added services initialization callback for client entry points
- Refactored and added new tests
- Bug fixing
- Updated documentation

### DIFF
--- a/uberfire-security/uberfire-security-management/uberfire-security-management-backend/src/main/java/org/uberfire/ext/security/management/UberfireRoleManager.java
+++ b/uberfire-security/uberfire-security-management/uberfire-security-management-backend/src/main/java/org/uberfire/ext/security/management/UberfireRoleManager.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *  
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *  
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.uberfire.ext.security.management;
+
+import org.jboss.errai.security.shared.api.Role;
+import org.uberfire.ext.security.management.api.*;
+import org.uberfire.ext.security.management.api.exception.GroupNotFoundException;
+import org.uberfire.ext.security.management.api.exception.SecurityManagementException;
+import org.uberfire.ext.security.management.search.RolesRuntimeSearchEngine;
+import org.uberfire.ext.security.management.search.RuntimeSearchEngine;
+import org.uberfire.ext.security.management.util.SecurityManagementUtils;
+
+import javax.enterprise.context.Dependent;
+import javax.inject.Named;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+/**
+ * <p>The default role manager implementation for UF based applications.</p>
+ * <p>Roles are not coming from any external system, are just the ones registered in the application</p>
+ * <p>CRUD operations for roles are not allowed, only search is allowed.</p>
+ * 
+ * @since 0.8.0
+ */
+@Dependent
+@Named("uberfireRoleManager")
+public class UberfireRoleManager implements RoleManager, ContextualManager {
+
+    UserSystemManager userSystemManager;
+    RuntimeSearchEngine<Role> rolesSearchEngine;
+    
+    @Override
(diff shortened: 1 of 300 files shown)


romartin:
After this PR has been merged and updated artifacts available at nexus, please merge this one for the wb integration -> https://github.com/droolsjbpm/kie-wb-distributions/pull/172

uberfire-security/uberfire-security-management/uberfire-security-management-backend/src/main/java/org/uberfire/ext/security/management/UberfireRoleManager.java
```diff
@@ -0,0 +1,106 @@
...
+
+    // Check that roles registered in the framework are created as realm groups as well, if not they cannot be used, as it couldn't be assigned.
+    protected Set<Role> getRegisteredRoles() {
+        final Set<Role> result = new LinkedHashSet<Role>();
+        final Set<Role> registeredRoles = SecurityManagementUtils.getRegisteredRoles();
+        if ( null != registeredRoles && !registeredRoles.isEmpty() ) {
+            GroupManager groupManager = userSystemManager.groups();
+            for (final Role registeredRole : registeredRoles) {
+                try {
+                    groupManager.get(registeredRole.getName());
```
csadilek: @romartin what is this get call doing here? The result is not being assigned. Is there a side-effect your relying on or is it just a left-over?

uberfire-security/uberfire-security-management/uberfire-security-management-wildfly/src/main/java/org/uberfire/ext/security/management/wildfly/properties/WildflyGroupPropertiesManager.java
```diff
@@ -101,24 +101,41 @@ public Group get(String identifier) throws SecurityManagementException {
...
                         for (final String name : groups) {
                             if (!allGroups.contains(name)) {
                                 String error = "Error getting groups for user. User's group '" + name + "' does not exist.";
                                 LOG.error(error);
                                 throw new SecurityManagementException(error);
                             }
-                            result.add(createGroup(name));
+                            
+                            if (registeredRoles.contains(name)) {
+                                // Is a role.
```
csadilek: Don't like that this code is duplicated. Can we move it to a base class or reuse it somehow?

csadilek:
Looks good, just two minor comments above.

uberfire-security/uberfire-security-management/uberfire-security-management-backend/src/main/java/org/uberfire/ext/security/management/UberfireRoleManager.java
romartin: This method is used as well. Let me explain, when you want to assign roles to a user, you open the roles editor, which uses this role manager to obtain the system roles to show in the list for selection.

For the Tomcat and Keycloak, this method check that the roles "registered" (so in the RolesRegistry) do exist in the realm, but on the wildfly/eap case, which use the realms based on properties, registered roles could not be registered in the app-roles.properties, as this file does contain the roles, it contains only role assignments, so if a registered role is not in this file, it does not mean it doesn't exist, just that it has not been assigned yet. 

So for the wildfly/eap one this methods returns all registered roles, and for the other providers, the registered roles than exist in the realm (if the role does not exist, it cannot be assigned :)  )

romartin: and the method is used in the above "search" one

csadilek: @romartin I was just talking about line 95 which has a call to groupManager.get without assigning the result. I understood what the method's purpose was ;).

romartin: ahh @csadilek I undestand :) Yes, the result it's not assigned because i don't need it :) I mean, if the group not exist, an exception will be thrown, so i'm just adding the role in the list if there is no exception, so the group/role exist on the realm.

romartin: so this "groupManager.get(..)" throws an exception is not found

romartin: I will add a comment on it to make it clear

romartin:
@csadilek wait please! I found another code block duplicate, just creating new commit...

romartin:
@csadilek fixed, PR ready, thx!

csadilek:
+1

csadilek:
master: b78badd97a1731e7bc054e15237e3f821e51bcd7

PR #174, opened 2016-01-14
GUVNOR-2338: Copy/Rename modal doesn't disappear when used from Project Explorer links

paulovmr:


manstis:
It looks like you might have implemented the "inverse" of what I believe the requirement to be!?!

When you use a CopyPopup or RenamePopup (or DeletePopup) it should be hidden *on success* and remain visible if there was an error (e.g. validation error) no matter where the popups are shown (Editor's menu bar, Project Explorer "technical view", Project Explorer "business view" etc)

paulovmr:
@manstis Yes, looks like that, but no. :)

In one of my previous PRs (https://github.com/uberfire/uberfire-extensions/pull/163), I mistakenly removed the hide() call from the onSuccess() method, and that caused this issue. I did that because there is a case where even if the validator succeeded the popup must remain open (when BasicMenuFileImpl uses the CopyPopup).

Now, I've returned the hide() call to its place, but it will not be triggered only when it is the BasicMenuFileImpl class that is using the CopyPopup (this is what I should have done previously, to not break the behavior for any other users of the CopyPopup, such as the ProjectMenu clas, on kie-wb-commons).

paulovmr:
One more thing, the bug happens with one of this "other users" of the CopyPopup (or RenamePopup), and not with the BasicMenuFileImpl class.

paulovmr:
To better explain the case where the popup should remain visible, it happens when the validator succeeds and the call to copy service fails. Then a inline error must be shown in the popup (https://github.com/paulovmr/uberfire-extensions/blob/158c6bfa8337caffa75e47ff47fd318907b81e42/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/menu/BasicFileMenuBuilderImpl.java#L285).

manstis:
OK, is not then a case of adding ```view.hide()``` to https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/widgets/BaseViewPresenter.java#L304 as is done here https://github.com/paulovmr/uberfire-extensions/blob/158c6bfa8337caffa75e47ff47fd318907b81e42/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/menu/BasicFileMenuBuilderImpl.java#L271

paulovmr:
@manstis Just clarifying a little more about why I decided to do it that way: the BaseViewPresenter is only one place where the CopyPopup/RenamePopup classes are instantiated. In the kie-wb-commons repo there are another three places doing that, and could exist more in other repos. I thought that by creating the preventHideOnSuccess() method, I could maintain an expected behavior on the popups and make possible to override this behavior if necessary.

Anyway, I think both approaches are valid. :)

Which one do you think we should choose?

manstis:
@paulovmr Hi.. I'm not trying to be troublesome :) 

Personally, and it is my personal opinion, I think it should be the code instantiating these popups to decide when to close them; the "consuming" code retains responsibility for when to show the popup and when to hide it. 

This would mean having ```view.hide()``` in one of the instantiating code's callbacks - whether it's the ```Validator.onSuccess(..)``` or ```Validator.onFailure(...)``` or ```CommandWithFileNameAndCommitMessage.execute()``` or.. so forth.

My opinion is however one of many; so let's see what the others think..

@csadilek @porcelli @Rikkola @wmedvede Any thoughts?

Rikkola:
Personally I would vote for not to have this in the popup, but where the popup is created.
The reason is that this method adds a boolean flag into the popup code and those tend to add unnecessary complexity. These popup classes are quite simple right now and I doubt they ever get too complex, but this is a rule I have tried to follow myself. It is usually better to call the methods to execute actions than add booleans to store states. Helps with readability, testing and so on.

Just my thoughts on a Friday evening :beers:

paulovmr:
@Rikkola and @manstis good points :)

I will change the approach then. Thanks again! :)

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/menu/BasicFileMenuBuilderImpl.java
```diff
@@ -207,6 +207,7 @@ public boolean error( final Message message,
                     return false;
                 }
 
+                renamePopupView.hide();
```
manstis: For consistency with the ```show()``` we should call ```RenamePopup.hide()```.

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/menu/BasicFileMenuBuilderImpl.java
```diff
@@ -263,19 +264,19 @@ public void execute( final FileNameAndCommitMessage details ) {
...
     }
 
-    private RemoteCallback<Path> getCopySuccessCallback( final CopyPopupViewImpl copyView ) {
+    private RemoteCallback<Path> getCopySuccessCallback( final CopyPopupView copyPopupView ) {
         return new RemoteCallback<Path>() {
 
             @Override
             public void callback( final Path path ) {
-                copyView.hide();
+                copyPopupView.hide();
```
manstis: For consistency with the ```show()``` we should call ```CopyPopup.hide()```.

uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/menu/BasicFileMenuBuilderImpl.java
```diff
@@ -288,6 +289,7 @@ public boolean error( final Message message,
                     return false;
                 }
 
+                copyPopupView.hide();
```
manstis: For consistency with the ```show()``` we should call ```CopyPopup.hide()```.

manstis:
Other than the change to use the Presenter's ```hide()``` method I think this is much clearer to follow than the flag you had before. Thanks for making the changes... once you switch to using the Presenter I'll OK the merge.

paulovmr:
@manstis Hi, thanks for the review. :)

About calling the popup instead of the view, I thought about that too... But, the hide() method is always called inside a callback that is inside another callback that is passed in the popup instantiation. So, I don't have the popup yet.

There is a way to do this, but I don't know if it's better: create the popup without the callback, create the callback that uses the popup, and set this callback in the popup.

Do you see another better way?

manstis:
Hi @paulovmr You're right!

In which case I'll retract my comments.. I don't see a better (cleaner) approach :)

manstis:
+1 to merge.

csadilek:
master: e0f08e81cea117a134a9fe165d8e372dc9c24a22

PR #176, opened 2016-01-16
GUVNOR-2342: added tests

jomarko:


manstis:
Hello, license needs updating and IMO the test would be better if it verified after each view.set call - otherwise we're not verifying which call set the footer to which value.

jomarko:
@manstis you are right, I will fix it today.

manstis:
+1 to merge.

csadilek:
master: d64d8f6e51da915cc359685bc360ba6901646dfa

PR #185, opened 2016-01-28
Update to Errai 4.0.0-SNAPSHOT, GWT 2.8.0-beta1, and Java 8.

mbarkley:
* Switch guava version to match guava-gwt used by Errai.
* Update usage of old SyncBeanManager API.
* Remove @New (no longer supported).

### DIFF
--- a/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/BaseEditor.java
+++ b/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/BaseEditor.java
@@ -77,7 +77,6 @@ public abstract class BaseEditor {
     protected VersionRecordManager versionRecordManager;
 
     @Inject
-    @New
     protected BasicFileMenuBuilder menuBuilder;
 
     @Inject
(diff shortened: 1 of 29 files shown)


csadilek:
+1

csadilek:
master: b1589f32dfe5eea6bf68b0e28245689f6729c259

manstis:
IIRC there was a heap of trouble with guava and other versions brought in by either Drools itself, or one of the dependencies Drools brought in for Maven support or Aether... I forget which... just a cautionary notice for when these changes cascade to KIE Workbench...

csadilek:
@manstis we've worked on this for weeks now and tested it the last 3 days. We've done our best to spot problems (found and fixed many). We're in the middle of upgrading all repositories. At this point, the only way to find more problems is to continue with this upgrade.

manstis:
@csadilek That I do not doubt and this is the way forward... I just recall there having been problems elsewhere in the past and mention for information only :-)

psiroky:
Just FYI: @manstis is right, there are certain issues with Drools Maven/Aether/Guava, see https://issues.jboss.org/browse/DROOLS-1045. I will be looking into this from Drools point and will make sure it works wit the newer Guava version (it will likely mean upgrading Maven and Aether deps as well).

csadilek:
@manstis @psiroky I don't doubt it. I guess we'll have to keep an eye on this in the future. We are happy to help with related issues as they come up.

psiroky:
This is actually something we would be looking into anyway -- we really need to get rid of very old Guava 13. So no worries, I believe we should be able to fix this with few component upgrades.

csadilek:
OK great. Even if we can't upgrade guava in drools it shouldn't be a blocker as we only need the newest guava for GWT compilation. We can keep using the old one on the server if we need to.

PR #189, opened 2016-02-02
Fix bean manager use so that only a single PerspectiveEditorActivity …

mbarkley:
…can be looked up.

In the previous usage, IOC.getBeanManager().lookupBeans(PerspectiveEditorActivity.class)
would return 3 bean defs with a reference to the same activity.


ederign:
+1

csadilek:
master: 5f12cfc33db70628b3c0c9aa1db8863821eac4c3

PR #190, opened 2016-02-02
@ApplicationScoped becomes lazy. We should transform this in @EntryPo…

ederign:
…int to becomes eager


csadilek:
+1

csadilek:
master: 532ee2035cc5c1a883ed997da4117aefe2872fae

PR #199, opened 2016-02-09
Keycloak 1.8 integration

romartin:
This PR add support for latest Keylocak 1.8.0.Final. This is the PR for 0.8.x branch, as it hurries for the kie 6.3.x build and qe testing.  I will send the PR for master branch in a while.
Thanks!

### DIFF
--- a/uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/BaseKeyCloakManager.java
+++ b/uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/BaseKeyCloakManager.java
@@ -101,13 +101,21 @@ public abstract class BaseKeyCloakManager {
         LOG.debug("Configuration of KeyCloak provider finished.");
     }
 
-    protected synchronized Keycloak getKeyCloakInstance() {
+    protected Keycloak getKeyCloakInstance() {
         if (this.keycloak == null) {
             this.keycloak = Keycloak.getInstance(authServer, realm, user, password, clientId, clientPassword);
         }
         return keycloak;
     }
-    
+
+    protected boolean close() {
+        if (null != this.keycloak) {
+            keycloak.close();
+            return true;
+        }
+        return false;
+    }
+
     protected RealmResource getRealmResource() {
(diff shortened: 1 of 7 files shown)


romartin:
NOTE: There are no changes on the kie-wb-distributions repository expected for this update. Only uf-extensions is affected.

uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/BaseKeyCloakManager.java
```diff
@@ -102,12 +102,20 @@ protected void loadConfig( final ConfigProperties config ) {
     }
 
     protected synchronized Keycloak getKeyCloakInstance() {
-        if (this.keycloak == null) {
+        if ( null == this.keycloak ) {
```
psiroky: Was it intended? Looks a bit weird :)

csadilek: Yes, looks like a left-over, @romartin? Also can you add some JavaDoc why this method is synchronized?

csadilek:
@romartin just some minor things, see comments above. Then good to go.

uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/BaseKeyCloakManager.java
romartin: will remove the syncronization here yes, does not makes sense. About the notation of the if... just notation, no matter for me to revert it :)

romartin:
@psiroky @csadilek fixed. thx!

csadilek:
@romartin since reference assignments are atomic, the worst that could happen to you after removing the synchronized block is that Keycloak.getInstance is invoked multiple times (and the field is getting reassigned). If you can live with that (seems ok to me), removing synchronized is fine. Can you please comment.

romartin:
@csadilek @psiroky Closing for now. Let's see which concrete version is required to be present. Thx, will keep u updated

PR #201, opened 2016-02-10
i18n BZs: 1284792 1284799 1284816 12848818

ederign:


manstis:
Way to go Eder! +1 to merge.

ederign:
@manstis btw on 1284799 the screen title will remain the same
on 1284816 the place name will remain in english because is a uberfire concept.

is that right?

manstis:
"Place Names" as in PlaceManager.goto("placeName") don't need to be i18n'ed as they are internal logical identifiers for physical screens and are not (should not) be exposed to Users. If we are exposing them to Users we should probably change to use a "regular" GWT PropertyBundle constant (ie like all other i18n).. Have I understood correctly?

ederign:
I mean, On the edit screen popup there is a label called Place Name and a list with all the available screens/editors.

I don't think that translates this is a good idea because that is related with uberfire screen/editor identifier concept.

ederign:
I've changed my mind! I'll i18n this! :)

csadilek:
master: 5f9e274b7c6f016e75eaf480d7d18c8efa3f56af

PR #212, opened 2016-02-17
Translations pulled from Zanata (0.8.x)

cristianonicolai:


manstis:
+1 to merge.

(Michael B usually does this for all repositories when performing a build, so unless you have good reason to want to check i18n before a release you don't need do this - word of caution *if* you do this you should also run the "post fetch" updates such as mvn replacer:replace - *some* i18n eg French contain single apostrophes which need escaping to double - single apostrophes for GWT....)

cristianonicolai:
@manstis hm thanks, I haven' t done that replacer so maybe we should hold this. The reason why I pulled this is because the number of i18n that we're trying to get verified. Getting the updated content from Zanata helps finding out if it is an issue with translation or code.

manstis:
There's no problem with you doing it if you need ti check i18n before a release, just be sure to run mvn replacer:replace in the repository root before committing :-) IIRC it's covered in the README in droolsjbpm-build-bootstrap... but real developers don't read that!!! --- me included ;-)

cristianonicolai:
@manstis thanks heaps! got it sorted now and also checked the docs :)

manstis:
+1 to merge.

PR #214, opened 2016-02-19
BZ1304659: File Explorer: Package is expanded only after second click

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1304659

### DIFF
--- a/uberfire-widgets/uberfire-widgets-core/uberfire-widgets-core-client/src/main/java/org/uberfire/ext/widgets/core/client/tree/TreeItem.java
+++ b/uberfire-widgets/uberfire-widgets-core/uberfire-widgets-core-client/src/main/java/org/uberfire/ext/widgets/core/client/tree/TreeItem.java
@@ -18,6 +18,7 @@ package org.uberfire.ext.widgets.core.client.tree;
 
 import java.util.Iterator;
 
+import com.google.gwt.core.client.GWT;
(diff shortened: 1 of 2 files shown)


ederign:
@manstis how this solve the issue?

uberfire-widgets/uberfire-widgets-core/uberfire-widgets-core-client/src/main/java/org/uberfire/ext/widgets/core/client/tree/TreeItem.java
```diff
@@ -87,7 +88,6 @@ public TreeItem( final Type type,
                         public void onClick( ClickEvent event ) {
                             if ( !isSelected ) {
                                 updateSelected();
-                                return;
```
manstis: @ederign Here solves the issue.. previously when the DOM element was clicked it first selected it and exited; now it selects it and opens it.

ederign:
+1!

uberfire-widgets/uberfire-widgets-core/uberfire-widgets-core-client/src/main/java/org/uberfire/ext/widgets/core/client/tree/TreeItem.java
ederign: thanks!

PR #215, opened 2016-02-19
GUVNOR-2430: Delete popup: Remove confirmation check

manstis:
See https://issues.jboss.org/browse/GUVNOR-2430

(No tests as the popups are to be rewritten for https://issues.jboss.org/browse/GUVNOR-2431)


ederign:
+1!

PR #218, opened 2016-02-23
GUVNOR-2414: When deleting a file from a project, an error message is shown even though the file is succesfully deleted.

paulovmr:
This PR adds (for uberfire-extensions' users) the capability to implement restrictors to delete, rename and copy services. 

The code added on deleteIfExists, renameIfExists and copyIfExists methods (in their respective services) were migrated from kie-wb-common.

### DIFF
--- a/uberfire-commons-editor/uberfire-commons-editor-api/src/main/java/org/uberfire/ext/editor/commons/service/CopyService.java
+++ b/uberfire-commons-editor/uberfire-commons-editor-api/src/main/java/org/uberfire/ext/editor/commons/service/CopyService.java
@@ -15,10 +15,32 @@
  */
 package org.uberfire.ext.editor.commons.service;
 
+import java.util.Collection;
+
 import org.jboss.errai.bus.server.annotations.Remote;
+import org.uberfire.backend.vfs.Path;
 import org.uberfire.ext.editor.commons.service.support.SupportsCopy;
 
+/**
+ * Responsible for paths copy.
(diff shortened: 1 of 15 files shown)


uberfire-commons-editor/uberfire-commons-editor-api/src/main/java/org/uberfire/ext/editor/commons/service/CopyService.java
```diff
@@ -15,10 +15,32 @@
...
+     * @param comment Comment about the renaming.
+     */
+    void copyIfExists( final Collection<Path> paths,
+                       final String newName,
+                       final String comment );
+
+    /**
+     * Verifies if a path can be copied.
+     * @param path Path to be verified.
+     * @return true if the path can be copied and false otherwise.
```
psiroky: This looks a bit weird considering the method name. If the method returns `true`, the name suggest is has restriction, so I would assume it should not be copied?

paulovmr: @psiroky Yes, the comment is wrong (for the other services too). I will fix it. Thanks.

ederign:
@manstis as I've helped Paulo with the solution, would you mind to review this?

manstis:
@paulovmr Great stuff!

+1 to merge.

PR #219, opened 2016-02-23
GUVNOR-2414: When deleting a file from a project, an error message is shown even though the file is succesfully deleted.

paulovmr:
This PR adds (for uberfire-extensions' users) the capability to implement restrictors to delete, rename and copy services. 

The code added on deleteIfExists, renameIfExists and copyIfExists methods (in their respective services) were migrated from kie-wb-common.


manstis:
@paulovmr Great stuff!

+1 to merge.

PR #221, opened 2016-03-02
GUVNOR-2360 - Data Objects cannot be copied in the Authoring perspective

wmedvede:
- FileNameValidators expects the filename + extension.

### DIFF
--- a/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopup.java
+++ b/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopup.java
@@ -94,7 +94,7 @@ public class CopyPopup implements CopyPopupView.Presenter {
                 ? originalFileName.substring( originalFileName.lastIndexOf( "." ) ) : "" );
         final String fileName = baseFileName + extension;
 
-        validator.validate( baseFileName,
+        validator.validate( fileName,
                             new ValidatorWithReasonCallback() {
                                 @Override
                                 public void onFailure( final String reason ) {
--- a/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/RenamePopup.java
+++ b/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/RenamePopup.java
@@ -94,7 +94,7 @@ public class RenamePopup implements RenamePopupView.Presenter {
                 ? originalFileName.substring( originalFileName.lastIndexOf( "." ) ) : "" );
         final String fileName = baseFileName + extension;
 
-        validator.validate( baseFileName,
+        validator.validate( fileName,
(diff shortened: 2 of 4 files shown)


manstis:
+1 to merge.

wmedvede:
@paulovmr 
Hello Paulo the following PR contains the fix the the small issue we chatted about with the CopyPopup. I saw that the same happens with the RenamePopup, so I've just fixed both. Tests remains the same, since the issue is that FileNameValidators expects the fileName + extensions.

If you can review it will be great.

manstis:
@wmedvede Only one file changed in this PR (not ```RenamePopup```)

manstis:
BTW I saw your chat on IRC.. so seeing the PR thought I'd +1 it..

wmedvede:
@manstis ups, seems like I've forgot to add the RenamePopup

wmedvede:
@manstis 

Both Copy and Rename popups added!
Ready to merge.

csadilek:
@wmedvede @manstis -1, this is missing tests or better updates of the existing tests to verify that filename + extension is passed to the validator, not just any string.

manstis:
@csadilek I was stood in the rain waiting for my son to come out of school and was thinking the exact same thing.. the JIRA was caused by a regression, so we should test for the same regression in the future.. should be a simple change.

paulovmr:
@wmedvede Hi Walter, I agree with the change. Also, it will be necessary to change the following method, to remove the file extension in the "name" parameter, so the method works as before: https://github.com/uberfire/uberfire-extensions/blob/master/uberfire-runtime-plugins/uberfire-runtime-plugins-client/src/main/java/org/uberfire/ext/plugin/client/validation/PluginNameValidator.java#L78

wmedvede:
@paulovmr  Hi Paulo, I'm not sure what you mean with "remove the extension" in the case of the plugins.
In the case of CopyPopup and RenamePopup the problem was that the popus removed the extension, but the file validators need the extension to perform the validation.
On the other hand If I create, copy, and rename a plugin it seems to work. So I don't know exactly how this affects the plugins.  ¿?

paulovmr:
@wmedvede Yes, the validation still works with your change, but only because of the server-side, and only in some cases. In the client-side, this method that I mentioned makes the validation looking to the name of the registered plugins, and they are different from these plugins file names (and paths).

The following comparison would never be true if the "name" contains the file extension: https://github.com/uberfire/uberfire-extensions/blob/master/uberfire-runtime-plugins/uberfire-runtime-plugins-client/src/main/java/org/uberfire/ext/plugin/client/validation/PluginNameValidator.java#L92

wmedvede:
@manstis @paulovmr 

Hello guys, according to what we chatted the following PR reverts the regression introduced by the plugins stuff. So I guess this PR can now be merged. Finally the following https://issues.jboss.org/browse/UF-284 was opened for investigation.

paulovmr:
+1 to merge. I will work on this new ticket.

csadilek:
0.8.x: e374ee36a69205f96eddfd03bed434495f2d311c

I (by mistake) pushed this to master as well. So, no need to send another PR :).

master: 4b7ae2187412d81a77658c67c3999d5ac17df0f8

PR #222, opened 2016-03-03
GUVNOR-2452: improvements on YesNoCancellPopup in order to be more fl…

porcelli:
…exible in terms of size


ederign:
+1. Optionally I would add tests for clearScrollHeight() method.

porcelli:
Don't think tests would bring anything in this particular case, as to test it you'll have to mock the only possible side effect (the Element())

ederign:
I would mock scroll panel, and make sure that after a YesNoCancelPopup.clearScrollHeight(), clear height for the panel was called. But again, optionally, not necessary. +1 anyway

csadilek:
0.8.x: 9d95185d710600888710f0715e24a2d9cc9d595b

PR #228, opened 2016-03-07
UF-282: Removed cancel button from change password popup

romartin:


manstis:
Personally I dislike methods taking null parameter values... and would prefer to have seen an overloaded method.. but still; there's nothing technically wrong here.. so +1 to merge.

romartin:
@manstis tbh I'm agree, lemme do it in a while :) please wait for the merge just a bit

romartin:
@manstis hmmm checking again, I'm missing something sorry...what do you mean? the public "show" method is the overloaded one, providing different public methods with arguments, the nulls are used just internally on the different overladed methods to rely centralized onthe call to the method with all the arguments

manstis:
@romartin Yeah, I guess I'd have elected for adding the overloads to the ```View``` too - keeping the ```View```s implementation black-boxed - i.e. use of ```YesNoCancelPopup```... not going to argue about it... happy to +1 (just because it's different to what I'd have done, doesn't make it wrong)

manstis:
+1 to merge.

romartin:
@manstis ahh for the view! got it sorry ;) makes sense as well, good point, I prefer it too, refactoring

romartin:
@manstis done ;) thx!

manstis:
+1 to merge. Thanks, together we'll make this world a better place ;-)

csadilek:
master: 18360c1ecfec885b93d383778bbe8219e299a3f2

romartin:
@manstis hahaha xDD

PR #241, opened 2016-03-12
UF-287: User and group management - New admin client for the Keycloak provider.

romartin:


manstis:
I don't understand the changes but I notice CustomClientTest appears to be completely commented out. Is this correct? Also I noticed a *final* constant being localhost:8080/auth (not in a test). Is this correct?

romartin:
@manstis Ohhh yes argghh forget that local test :( removing now. 

Btw, why you say do don't understand the changes? Lemme explain, the issue with the current KC provider was that it was using the KC admin client, which is not supported on EAP 6.x, as the client relies on JAXRS API 2.X. On the other hand, is not clear the further support for the admin client from KC side, it's neither productized. So the only way to have the KC provider running again on all the environments was just creating a new custom client. This impl has just added  a rest easy client, the provider code itself is quite unmodified... Does this answers your question? :)

romartin:
@manstis about the constant with the admin URL... it's a default value, not i18n... what do you propose?

romartin:
Custom test removed. Sorry!

manstis:
I don't know KC  or it's integration (does anyone other than you?!?) so I'm blind reviewing the PR :-( As for the constant URL would it need to be different if kie-workbench is deployed not on port 8080 (or localhost)?

romartin:
@manstis hehe np Mike! There is no requirement for it but I was feeling that the job was not finished with the KC provider... So I commented about this to managers, as I said, feel free just to drop this is any other approach will be used in the future, for me np, at least I have learned and remembered about remote clients stuff, so great! :) Last decision was that they agreed on merging this into master as tech preview.

About the URL - This URL is intended to be a setting present in the configuration file specified by the user, this constant is its  default value. So it the server is on different/host it's documented that this URL have to be specified as a settings. The most common use case is specifying it in the settings always (appears all docs). So if you think so, the only thing we can do is just making the setting mandatory and remove the constant from the code, np! just let me know :)

About the PR... I know it's not under you total control that new stuff hehe, but I have been careful adding all tests, copyrights, comments, javadocs... so hope that could be fine :)

Thanks for the review! appreciated!

uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/README.md
```diff
@@ -17,114 +31,65 @@ Installation notes
 Step 1 - Dependencies
 ---------------------
 
-This provider uses the *KeyCloak Admin Client* (which uses KeyCloak's REST API services) to manage the different entities, so the following keycloak artifacts are required:                         
+This provider uses a custom RestEasy client that consumes the Keycloak remote admin services, so the following JAXRS, RestEasy
```
csadilek: The readme needs an update here. It's not clear which provider is meant here ("this provider"). I can infer that it's likely the credentials provider, but it should be made clear with a heading or subsection.

uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/BaseClientFactory.java
```diff
@@ -0,0 +1,47 @@
...
+
+package org.uberfire.ext.security.management.keycloak;
+
+import org.uberfire.commons.config.ConfigProperties;
+import org.uberfire.ext.security.management.keycloak.client.ClientFactory;
+import org.uberfire.ext.security.management.keycloak.client.Keycloak;
+
+/**
+ *  @since 0.9.0
+ */
```
csadilek: Wouldn't mind some Java docs other than ```@since```.

uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/CredentialsClientFactory.java
```diff
@@ -0,0 +1,71 @@
...
+    public void init(final ConfigProperties config) {
+        final ConfigProperties.ConfigProperty authServer = config.get("org.uberfire.ext.security.management.keycloak.authServer", DEFAULT_AUTH_SERVER);
+        final ConfigProperties.ConfigProperty realm = config.get("org.uberfire.ext.security.management.keycloak.realm", DEFAULT_REALM);
+        final ConfigProperties.ConfigProperty user = config.get("org.uberfire.ext.security.management.keycloak.user", DEFAULT_USER);
+        final ConfigProperties.ConfigProperty password = config.get("org.uberfire.ext.security.management.keycloak.password", DEFAULT_PASSWORD);
+        final ConfigProperties.ConfigProperty clientId = config.get("org.uberfire.ext.security.management.keycloak.clientId", DEFAULT_CLIENT_ID);
+        final ConfigProperties.ConfigProperty clientSecret = config.get("org.uberfire.ext.security.management.keycloak.clientSecret", DEFAULT_CLIENT_SECRET);
+
+        // Check mandatory properties.
+        if (!isConfigPropertySet(authServer))
```
csadilek: This would be a good API enhancement to UberFire. You're passing default values to .get in case the property has not been defined but really you want to fail instead. So, adding a config getOrFail("org.uberfire.ext.security.management.keycloak.authServer") would save you a lot of code duplication here. /cc @ederign

uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/client/auth/credentials/AuthTokenManager.java
```diff
@@ -0,0 +1,125 @@
...
+        try {
+            
+            TokenService client = createTokenService();
+            AccessTokenResponse response = client.grantToken(config.getRealm(), authorization, mvm);
+
+            int requestTime = Time.currentTime();
+            expirationTime = requestTime + response.getExpiresIn();
+            this.accessTokenResponse = response;
+            
+        } catch (BadRequestException e) {
```
csadilek: What's the reason for catching a BadRequestException here? Are there versions that don't support this refresh token request in which case you trying grantToken? If so, please add a comment.

csadilek:
@romartin looks good (especially the tests added). Just a few minor comments inline.

romartin:
@manstis @csadilek thanks guys! Agree with the comments, will work on them later and provide a new commit. thx!

romartin:
@csadilek hi! added a new commit, please merge this one before:
https://github.com/uberfire/uberfire/pull/323
Thanks!

uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/README.md
romartin: Yeah those deps apply for both impls, so I have added the comment that it applies for both

uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/CredentialsClientFactory.java
romartin: https://github.com/uberfire/uberfire/pull/323

uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/client/auth/credentials/AuthTokenManager.java
romartin: Added comment for it. Basically the idea is that is refresh token op fails, let's try to grant a new token again from scratch (not refreshing the current one)

romartin:
@csadilek commited again using only config.get(name, defaultValue) as commented. Thx

uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/client/auth/credentials/AuthTokenManager.java
csadilek: @romartin well, yes, but the question is what "fails" means in this case. You would only get a BadRequestException (HTTP 400) if the server decided that the request was invalid i.e. malformed (syntactically wrong)! Hence my question above about different versions. If you want to catch all kinds of errors you either need to check the status code or work with more mapped exceptions.

romartin: @csadilek well tbh I don't know exactly all kind of errors that this concrete endpoint can return, I just used same approach as the current admin client from KC team, which is working fine. So what do you propose? Look carefully to different errors types here and take different decisions?

csadilek: @romartin to say somebody else is doing it as well isn't the best argument ;). Like I wrote, you will only get a BadRequestException if the request is syntactically wrong which based on your description isn't what you want. You would have to make sure you get the right status code and if not fall back to ```grantToken```. You can use RestEasy's ClientResponse type for that. It's important to know what the REST endpoint you're invoking is actually returning in case of problems. You won't catch those problems by simply catching BadRequestException as this is only for malformed requests (HTTP 400 indicates a client error).

romartin: @csadilek hehe I know :) ok np, let me review this more in deep when possible... thx!

romartin:
@csadilek Added a new commit to handle the different client response errors when using the token service and a test for it. Thanks!

romartin:
@csadilek Hey! nullified the response after an error, as commented. Commited. Thx

csadilek:
+1

csadilek:
master: d4b50614351da3242930afa98831211571163e21

PR #244, opened 2016-03-15
BZ-1245729 - REST API access using a browser invalidates the session

mswiderski:
still trying to figure out how to write a test for this so feel free to hold on with merge

### DIFF
--- a/uberfire-security/uberfire-servlet-security/src/main/java/org/uberfire/ext/security/server/BasicAuthSecurityFilter.java
+++ b/uberfire-security/uberfire-servlet-security/src/main/java/org/uberfire/ext/security/server/BasicAuthSecurityFilter.java
@@ -38,10 +38,18 @@ public class BasicAuthSecurityFilter implements Filter {
 
     public static final String REALM_NAME_PARAM = "realmName";
 
-    @Inject
     private AuthenticationService authenticationService;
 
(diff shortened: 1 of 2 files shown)


mswiderski:
tests added so it's ready from my point of view

csadilek:
@mswiderski changes and test look good, but I am wondering why we are invalidating the session here at all? Do we want every request to /rest /maven2 /ws to re-authenticate?

csadilek:
Jenkins please retest this.

mswiderski:
@csadilek main reason is that if we won't do it we will ran into memory issues. They will be caused by session objects (such as User/Roles information not not limited to them) that are session scoped cdi beans or anything else that is stored in http session. Since these will be bound to http session that when accessed over rest (standalone client) will never be accessed again (same session) and then such http session will be kept open until session expires - times out. in workbench we have rather high value for it so when there are lots of REST api requests in short amount of time it will quickly cause server to run out of memory.

csadilek:
@mswiderski thank you, understood. I need to find out where the 24h session timeout in the workbench comes from. Always seemed way too high for me.

csadilek:
+1

csadilek:
master: acb52521f7d9b58ab1ec6a5dca10a391e73b1dac

PR #246, opened 2016-03-16
BZ-1245729 - REST API access using a browser invalidates the session

mswiderski:
(cherry picked from commit 57bb17f3b60fdabd2dec9c6d36304c1b6f212193)

back port of #244


csadilek:
0.8.x: 7fb0064115f7d8b59abc66327d9a84a98222e009

csadilek:
@mswiderski hm this actually doesn't compile on 0.8.x. Shouldn't have merged / thought I ran the build successfully but didn't apparently. I will send a PR with a fix.

mswiderski:
interesting, wound this be java version difference explaining why it worked fine on master?

thanks for fixing!

csadilek:
yes, it was just because 0.8.x is compiled for Java < 8, and the sessionProvider needed to final because the "effectively final" trick in Java 8 can't be used ;).

PR #250, opened 2016-03-18
Fix injection point in BasicAuthSecurityFilter for tomcat7

csadilek:
@mswiderski @ederign recent change/refactoring switched to constructor injection in this filter which doesn't work on tomcat7. so, this is switching back to field injection.


mswiderski:
+1 and apologies for all this trouble :(

csadilek:
@mswiderski no worries, I blame CDI :)

csadilek:
master: aa0ac06185fcbcc10fc3b6af9a82c3682ca6bdf4

PR #251, opened 2016-03-18
Fix injection point in BasicAuthSecurityFilter for tomcat7

csadilek:
PR for 0.8.x:

@mswiderski @ederign recent change/refactoring switched to constructor injection in this filter which doesn't work on tomcat7. so, this is switching back to field injection.


csadilek:
0.8.x: 44bc7f8f5b7724b8011e80ffba8525c0596b19ae

PR #253, opened 2016-03-23

nmirasch:


ederign:
https://github.com/uberfire/uberfire-extensions/pull/254

nmirasch:
I close this PR because the Jira referenced is not marked as blocker

PR #254, opened 2016-03-23

nmirasch:

### DIFF
--- a/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/DatePicker.java
+++ b/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/DatePicker.java
@@ -15,14 +15,19 @@
  */
 package org.uberfire.ext.widgets.common.client.common;
 
+import java.util.Arrays;
 import java.util.Date;
+import java.util.HashSet;
+import java.util.Set;
 
+import com.google.gwt.core.client.GWT;
 import com.google.gwt.dom.client.Element;
 import com.google.gwt.editor.client.IsEditor;
 import com.google.gwt.editor.client.LeafValueEditor;
 import com.google.gwt.event.logical.shared.ValueChangeEvent;
 import com.google.gwt.event.logical.shared.ValueChangeHandler;
 import com.google.gwt.i18n.client.DateTimeFormat;
+import com.google.gwt.i18n.client.LocaleInfo;
 import com.google.gwt.user.client.Event;
 import com.google.gwt.user.client.ui.Composite;
 import com.google.gwt.user.client.ui.HasEnabled;
@@ -99,18 +104,29 @@ public class DatePicker extends Composite
 
     private final boolean allowEmptyValues;
 
-    private final org.gwtbootstrap3.extras.datepicker.client.ui.DatePicker datePicker = new org.gwtbootstrap3.extras.datepicker.client.ui.DatePicker();
+    private final org.gwtbootstrap3.extras.datepicker.client.ui.DatePicker datePicker;
 
     public DatePicker() {
         this( true );
     }
 
+    /**
+     * Basic constuctor of uberfire datePicker without setting the datepicker container and without setting the hide
+     * handler to manage the interaction with org.gwtbootstrap3.extras.datepicker.client.ui.DatePicker popup
+     * @param datePicker
+     */
+    public DatePicker(org.gwtbootstrap3.extras.datepicker.client.ui.DatePicker datePicker) {
+        this.datePicker = datePicker;
+        this.allowEmptyValues = true;
+    }
+
     public DatePicker( final boolean allowEmptyValues ) {
+        datePicker = GWT.create(org.gwtbootstrap3.extras.datepicker.client.ui.DatePicker.class);
         this.allowEmptyValues = allowEmptyValues;
         datePicker.setContainer( RootPanel.get() );
 
         datePicker.setAutoClose( true );
-        datePicker.setFormat( DatePickerFormatUtilities.convertToBS3DateFormat( gwtDateFormat ) );
+        setFormat(gwtDateFormat);
 
         //When the popup Date Picker component is hidden assert empty values
         datePicker.addHideHandler( new HideHandler() {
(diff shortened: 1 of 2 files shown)


uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/DatePicker.java
```diff
@@ -99,18 +100,29 @@
...
+     * handler to manage the interaction with org.gwtbootstrap3.extras.datepicker.client.ui.DatePicker popup
+     * @param datePicker
+     */
+    public DatePicker(org.gwtbootstrap3.extras.datepicker.client.ui.DatePicker datePicker) {
+        this.datePicker = datePicker;
+        this.allowEmptyValues = true;
+    }
+
     public DatePicker( final boolean allowEmptyValues ) {
+        datePicker = new org.gwtbootstrap3.extras.datepicker.client.ui.DatePicker();
```
ederign: Why you didn't use GWT.create here?

uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/DatePicker.java
```diff
@@ -238,12 +250,75 @@ public void setForceParse( final boolean forceParse ) {
...
-        datePicker.setFormat( DatePickerFormatUtilities.convertToBS3DateFormat( gwtDateFormat ) );
+        datePicker.setFormat(DatePickerFormatUtilities.convertToBS3DateFormat(this.gwtDateFormat));
+    }
+
+    /**
+     * Sets the org.gwtbootstrap3.extras.datepicker.client.ui.DatePicker language and week start day
+     * depending on the activeLocale passed as parameter.
+     * There are incompatibilities between DateTimeFormatInfoImpl and DatePickerResourceBundles for es and de locales.
+     * This function modify in that cases the dateFormat to avoid show the name of the months
+     * TODO review locales es and de showing the name of the month.
```
ederign: This still need to be implemented? Or this todo is a leftover?

uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/DatePicker.java
```diff
@@ -238,12 +250,75 @@ public void setForceParse( final boolean forceParse ) {
...
+     * @param activeLocale
+     * @param dateFormat
+     * @return the dateformat to be applied depending of locale. In case of es and de, avoiding show the name of month
+     */
+    private String setDatePickerLangProperties(String activeLocale, String dateFormat){
+        DatePickerLanguage datePickerLanguage = DatePickerLanguage.EN;
+        DatePickerDayOfWeek weekStart = DatePickerDayOfWeek.SUNDAY;
+        String format = dateFormat;
+        if (activeLocale.equals("es")) {
```
ederign: I would remove all this logic to a separated class/enum. Something that translates activeLocale to -> DataPickerLangConverter (we need a better name for that) a and then I can make this conversions (datePickerLanguage = DatePickerLanguage.FR - weekStart = DatePickerDayOfWeek.MONDAY;)

Also, I would split that method in two, one to set/convert dataPickerLanguage/WeekStart and other to changes formats if necessary.

uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/DatePicker.java
```diff
@@ -238,12 +250,75 @@ public void setForceParse( final boolean forceParse ) {
...
+            datePickerLanguage = DatePickerLanguage.ES;
```
manstis: Personally, I'd try to find another way of doing the conversion from ```activeLocale``` to ```DatePickerLanguage```.. if another locale is added for BxMS we'd need to remember to change this class. Did you consider ```DatePickerLanguage.valueOf(..)``` with some minimal manipulation of ```activeLocale```?

ederign:
@nmirasch I've made some comments. Could you please take a look on them?

ederign:
@manstis we should avoid mid air colisions in reviews

uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/DatePicker.java
nmirasch: Better valueOf(), thanks!

uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/DatePicker.java
nmirasch: I only leave the tweak for the locale "es" and "ge" to not to show the name of months

uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/DatePicker.java
manstis: @nmirasch Glad it's of use.. 

I'd recommend double-checking at runtime (rather than relying on the values you use in your Unit Tests) as for example GWT's ```LocaleInfo``` JavaDocs state "Returns the name of this locale, such as 'default', 'en_US', etc". It's therefore feasible that 'en_US' won't match anything in ```DatePickerLanguage```. 

You might want to strip the language from the region (see https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-wb-distribution-home/kie-wb-distribution-home-client/src/main/java/org/kie/workbench/common/screens/home/client/widgets/home/HomeImagesHelper.java#L63 as an example).

uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/DatePicker.java
nmirasch: this have to be a temporary fix to the problems of the ES and DE with the month names. At least to could work with that locales. It's because of that I let here the TODO.

nmirasch:
I know this tweak for the problematic locales it's not the best solution. But this is a temporary fix that allow keep working with that locales when the framework is configured to show the name of the months in the dates. 
We can include this fix or wait for a more definitively solution.
Thanks for reviewing!

nmirasch:
@manstis I added the control of the current lang in getLangFromLocale you suggested. Can you review that?
Thanks!

manstis:
@nmirasch Thanks.. +1 to merge.

ederign:
@nmirasch:

> I know this tweak for the problematic locales it's not the best solution. But this is a temporary fix that allow keep working with that locales when the framework is configured to show the name of the months in the dates. 
> We can include this fix or wait for a more definitively solution.

This fix is for master right? Why not work with the best/definitive solution?

nmirasch:
@ederign I think the same, we need a better/definitive solution. I'm going to close this PR and take some time to find that.
@manstis @ederign thanks for all!

PR #261, opened 2016-04-06
Keycloak provider docs fix.

romartin:
Following the instructions now I realized that it was missing a single dependency that must be added as well. Thx


manstis:
+1 to merge.

PR #265, opened 2016-04-13
PLANNER-529: Workbench: Solver Editor: copying or renaming solver configuration file creates text file

paulovmr:
@ge0ffrey Hi, could you review this based on our discussion?


ge0ffrey:
+1 the new testcase includes the corner cases I asked about.
Thanks Paulo!

Side note: I noticed that wb seems to have it's own set of resource types (ResourceTypeDefinition.java) vs the set of kie-api (ResourceType.java). I am not saying this is necessarily a bad thing (DRY might not apply as the set maybe differs) - just wondering if anyone knows the motivation behind this.

ederign:
@paulovmr thanks!

@manstis any idea about @ge0ffrey concern?

manstis:
@ederign @ge0ffrey They do look similar don't they? The primary reason is that KIE Workbench's ``ResourceType``` interface comes from Uberfire.. which has - and should have - zero dependencies on Drools. Now.. if you want to propose Uberfire depends on Drools - or vice-versa.. we can consolidate interfaces.., but it'll be a cold day in hell before anyone agrees to that (nor should they).

ederign:
@manstis thanks for the clarification!

ge0ffrey:
@manstis thanks for info - agreed (and I definitely agree that UF and Drools cannot depend on each other.)

PR #271, opened 2016-04-22
BZ-1328722: [ALL LANG] Word 'of' is unlocalized in each string 'x-x of x' under each table result list.

paulovmr:


manstis:
@paulovmr Super * thanks!

+1 to merge.

uberfire-widgets/uberfire-widgets-table/src/main/java/org/uberfire/ext/widgets/table/client/UberfireSimplePager.java
```diff
@@ -372,16 +373,16 @@ protected String createText() {
...
-                    + " of "
+                    + " " + of() + " "
                     + formatter.format( dataSize );
         }
         return formatter.format( pageStart )
                 + "-"
                 + formatter.format( endIndex )
-                + ( exact ? " of " : " of over " )
+                + ( exact ? " " + of() + " " : " " + of() + " over " )
                 + formatter.format( dataSize );
```
csadilek: @paulovmr @manstis shouldn't over be localized here as well?

manstis: @csadilek Yes, well spotted.. although the BZ only says "of" should be localized, I think we should show willing ;)

@paulovmr Sorry.. I missed that.

paulovmr: @csadilek Nice catch, I missed that too...

paulovmr: @manstis No problem :)

paulovmr:
@csadilek @manstis Ok, ready for review. Thanks!

csadilek:
+1

PR #273, opened 2016-04-22
Move DataGrid to uf-widgets-table in order to remove dependency on uf…

mbarkley:
…-wb-client-views-patternfly.

This change makes it possible to use the uf table widget without transitively depending
on ubefire-workbench-client.


mbarkley:
Hi @ederign,

This change continues the table refactoring you started. There was a lingering issue in LiveSpark because uberfire-widgets-table still depended on uberfire-workbench-client-views-patternfly, which transitively brought in uberfire-workbench-client.

The side-effect of bringing in uberfire-workbench-client is that the generated LiveSpark application loaded with a div overlay covering the whole app (preventing interaction with any buttons on the page).

With this change, the only uberfire dependency in widgets-table is ubefire-commons.

Cheers.

ederign:
+1
@mbarkley thanks!

PR #274, opened 2016-04-22
Move CheckboxCellImpl to uberfire-widgets-table.

mbarkley:
@ederign @csadilek This PR is also for the benefit of LiveSpark, which uses the CheckboxCellImpl to display boolean properties in generated apps.


csadilek:
+1

PR #276, opened 2016-04-25
- test showing issue with freeing the (lucene) lock on a repository

mrietveld:
@porcelli (and @csadilek?): this is the issue I pinged you on IRC about a while back. 

This is not to say that code needs to be changed -- I don't know enough about how the `IOService` is used in the workbench/Uberfire. 

But my main point here is that calling `dispose()` on the [IOService](https://github.com/uberfire/uberfire/blob/master/uberfire-io/src/main/java/org/uberfire/io/IOService.java) is not enough if you're going to use the `IOService` on the same repository (well -- only if you're also using indexing, which might not always be the case?)

Thanks.


mrietveld:
Pushed the wrong branch.. :/ Deleting this.

PR #283, opened 2016-07-19
[0.9.x][UF-364] uberfire-security-management: support for WFLY10/EAP7

psiroky:
This is the first quick&dirty (but working) proposal, mainly to get  feedback.

Few notes:
- I basically copied the content of `uberfire-security-management-wildfly` and then did few needed code fixes. There is a certain duplication. It's just few classes, but still. If we agree it is worth the effort, I will extract the common code into some third module, and reuse it in the two specific ones.
- since WFLY10 requires Java 8, this change also means we need Java 8 to build uberfire
- I kept the original name `uberfire-security-management-wildfly`. We could rename that to `*-wildfly8`, but that would mean breaking backwards compatibility. Not sure what is worse in this case - inconsistent naming or breaking compatibility.

@ederign, @dgutierr, @romartin please take a look.

### DIFF
--- a/uberfire-extensions-bom/pom.xml
+++ b/uberfire-extensions-bom/pom.xml
@@ -341,6 +341,19 @@
 
       <dependency>
         <groupId>org.uberfire</groupId>
+        <artifactId>uberfire-security-management-wildfly10</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>org.uberfire</groupId>
+        <artifactId>uberfire-security-management-wildfly10</artifactId>
+        <version>${project.version}</version>
+        <classifier>sources</classifier>
+      </dependency>
+
+      <dependency>
+        <groupId>org.uberfire</groupId>
         <artifactId>uberfire-security-management-tomcat</artifactId>
         <version>${project.version}</version>
       </dependency>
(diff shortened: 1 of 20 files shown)


romartin:
HI @psiroky ,  I have reviewed this locally, as you say, right now there is quite some duplicity but it can be refactored by a common module yes. So for me +1, sounds good! Thanks! :)

PS: Just a comment, not sure if needed but probably there need to refactor the uberfire-security-management-webapp module as well, in order to use latest wildfly10 provider, once finished.

psiroky:
Thanks for the comments @romartin.

I see the following options for the new modules:

1) (this is a backwards incompatible change)

```
 - uberfire-security-management-wildfly (would become packaging POM instead of jar)
     |- uberfire-security-management-shared (common, shared code)
     |- uberfire-security-management-wildfly8
     |- uberfire-security-managemetn-wildfly10
```

2) (this is less obvious/more confusing, but would be fully backwards compatible)

```
- uberfire-security-management-wildfly (support for wildfly8, depending on the -shared module)
- uberfire-security-management-wildfly-shared (common, shared code, moved from -wildfly module)
- uberfire-security-management-wildfly10
```

@romartin, @ederign what would you prefer?

romartin:
Hi @psiroky , I prefer the option 1) , it's the clearest one for further usage... as you say it's not backwards compatible, but the migration it's just on Maven side, so should be transparent to the final users is it? So for me +1 :)

psiroky:
Just in case anyone is wondering: I am working on this and will hopefully push the updated version later today.

psiroky:
Replaced by https://github.com/uberfire/uberfire-extensions/pull/284.

PR #294, opened 2016-09-04
latest Zanata translation changes in uberfire-extensions from 2016-09-04

kiereleaseuser:


mbiarnes:
same content as PR #292 - so this PR can be closed

PR #315, opened 2017-03-20
[RHBPMS-4574] hack to avoid logout issues with Keycloak

psiroky:


psiroky:
@ederign or @porcelli could you please take a look?

porcelli (approved):

porcelli:
@psiroky maybe send a PR to master?

psiroky:
@porcelli good point, I will do that so that the code is in sync (and we should be able to remove this hack in future once we use newer WildFly versions).

